PR #13, opened 2016-12-13
Read alarm's description from alarm API instead of alarm-definition API

dhague:
This allows us to show the alarm's description with template variables filled in by Monasca, and potentially avoids a call to the alarm-definition API.


auhlig:
LGTM

dhague:
Discussed with @edda - no obvious booboos.

PR #15, opened 2016-12-21
disable cost_control if catalog does not contain billing service

auhlig:
Will that work @edda ?


edda:
lgtm

auhlig:
Cool. Thanks @edda

PR #18, opened 2016-12-28
add keypair nav

Carthaca:
@edda : was this intentionally left out?


edda:
Yes, it wasn't included in the project services navigation because it's not a project specific service and I didn't want the user to mistakenly assume that a created keypair is specific to a particular project.

edda:
But this might be overthinking things. If you have missed the navigation entry I don't mind adding it.

Carthaca:
Just ignore this, you are right about it not being project specific

PR #19, opened 2016-12-28
[identity] add translation to role names

Carthaca:
for more clarity we label the role names. most important for role `admin` to be identified as 'Keystone Admininistrator'

Happy Fighting over the label names 😄 💥 

@edda: okay? 

@urfuwo FYI

image


edda:
Awesome +1

PR #21, opened 2017-01-04
[CCM-294] put user in target for policy check

Carthaca:
@andypf / @tlesmann : is it intended to be used this way?


Carthaca:
thx

PR #22, opened 2017-01-05
textual change. in case no floating IPs available.

bozinsky:
Users are presented a dialog to create a floating ips, when floating IPs exhausted in the project. 
When a user does that, the IP is not assigned. Quite miss-leading when using it.


andypf:
lgtm

PR #23, opened 2017-01-05
fill policy with domain id for technical users

Carthaca:
@andypf please have a look

### DIFF
--- a/plugins/identity/app/controllers/identity/domains/technical_users_controller.rb
+++ b/plugins/identity/app/controllers/identity/domains/technical_users_controller.rb
@@ -5,16 +5,16 @@ module Identity
     class TechnicalUsersController < ::DashboardController
 
       authorization_required only: [:index, :new, :create]
-      
+
       def index
         services.identity.users(domain_id:@scoped_domain_id, "name__startswith" => 'T' ).length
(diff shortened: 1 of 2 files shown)


andypf:
lgtm

plugins/identity/app/controllers/identity/domains/technical_users_controller.rb
```diff
@@ -4,17 +4,23 @@ module Identity
   module Domains
     class TechnicalUsersController < ::DashboardController
 
+      before_filter do
+        @policy_default_params ||= {}
+        @policy_default_params[:user] = { domain_id: @scoped_domain_id }
```
Carthaca: maybe `@policy_default_params[:user] = current_user` instead ?

andypf: no, current_user doesn't have the attribute scoped_domain_id
But there is domain_id and user_domain_id. So you could change the rule in policy

andypf: ah sorry. You are right
Use current_user

ruvr: why is the user relevant at all in the policy? shouldn't it just check for other cloud_admin or a admin role for the current domain?

Carthaca:
checking for domain_id not being nil is what we want here: i.e. user is domain admin

andypf:
👍

PR #25, opened 2017-01-05
remove unused extra policy_default_param

Carthaca:
I could create an instance without these lines and without having a special nova role

@andypf please have a look


andypf:
I can't see any uses of project in instance policy. So 👍  from me

PR #31, opened 2017-01-12
[dns] allow cloud_dns_admin to do zone update/delete at central place

Carthaca:
@andypf : please have a look

### DIFF
--- a/plugins/dns_service/app/controllers/dns_service/zones_controller.rb
+++ b/plugins/dns_service/app/controllers/dns_service/zones_controller.rb
@@ -44,12 +44,11 @@ module DnsService
     end
 
     def edit
-      @zone = services.dns_service.find_zone(params[:id])
     end
 
     def update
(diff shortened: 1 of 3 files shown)


plugins/dns_service/app/controllers/dns_service/zones_controller.rb
```diff
@@ -44,13 +44,16 @@ def create
...
 
     def edit
-      @zone = services.dns_service.find_zone(params[:id])
     end
 
     def update
-      @zone = services.dns_service.find_zone(params[:id])
-      @zone.attributes=params[:zone]
-      if @zone.save
+      update_attributes = {
```
andypf: Actually you don't need a mapping here. It is already done here: https://github.com/sapcc/elektra/blob/master/plugins/dns_service/app/models/dns_service/zone.rb#L12-L18

Carthaca:
thx for the hint

PR #33, opened 2017-01-16
[Readme] add name origin

Carthaca:
@edda ok with you or is there more to tell?


edda:
Adjusted it with the meaning of the name that made us choose it :)

PR #38, opened 2017-01-20
Dns transfer

andypf:


plugins/dns_service/config/policy.json
```diff
@@ -2,19 +2,24 @@
   "context_is_cloud_dns_admin": "role:cloud_dns_admin",
   "context_is_dns_admin": "role:dns_admin",
   "admin": "rule:context_is_cloud_dns_admin or is_admin:True",
+  "primary_zone": "%(zone.type)s==SECONDARY",
```
Carthaca: where do we need that rule?

PR #39, opened 2017-01-25
[compute] compute_cloud_admin can view all instances

Carthaca:
added pagination + owning project

@edda / @andypf : ok with you?

would it be good to be able to turn this on/off as cloud_compute_admin or adding a flag/warning that such view is not limited to the project I am currently scoped to? same question applies to existing (for zones, flavors, images, networks ...) and coming super admin views

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,15 +1,15 @@
 module ApplicationHelper
   def render_paginatable(items)
-    return if !items or items.length==0
-    if @pagination_enabled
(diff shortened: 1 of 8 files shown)


Carthaca:
removed the pagination numbers if there is nothing to paginate

but: paginating instances breaks the quota usage display 😢

Carthaca:
fixed usage

andypf:
lgtm

Carthaca:
thx

PR #43, opened 2017-02-01
handle friendly id after project update

andypf:


app/models/friendly_id_entry.rb
```diff
@@ -50,4 +50,50 @@ def self.find_or_create_entry(class_name,scope,key,name)
...
   end
 
-end
\ No newline at end of file
+  def self.update_project_entry(project)
+    if project
+      sql = [
+        "class_name=? and lower(key)=? and lower(scope)=?",
+        'Project',
+        project.id,
```
Carthaca: Travis points to some issue here `PG::SyntaxError: ERROR:  syntax error at or near "["
       LINE 1: ...ries" WHERE (class_name='Project' and lower(key)=#[Double "i...` at `WHERE (class_name='Project' and lower(key)=#[Double "identity_service_driver"] and lower(scope)=#[Double "identity_service_driver"]) `

PR #53, opened 2017-02-27
Resource management/reduce quota

hgw77:


plugins/resource_management/app/controllers/resource_management/domain_admin_controller.rb
```diff
@@ -68,6 +70,55 @@ def update
...
+          rescue ArgumentError => e
+            @resource.add_validation_error(:approved_quota, 'is invalid: ' + e.message)
+          end
+        end
+        
+        # save the new quota to database
+        if @resource.save
+          @area = @resource.service.to_sym
+          @area_services = ResourceManagement::ServiceConfig.in_area(@area).map(&:name)
+          prepare_data_for_resource_list(@area_services)
```
majewsky: Is a `render action` missing here?

plugins/resource_management/app/controllers/resource_management/project_resources_controller.rb
```diff
@@ -21,7 +23,59 @@ def index
       @nearly_full_resources = @all_resources.where("usage <= approved_quota AND current_quota <= approved_quota AND usage >= 0.8 * approved_quota AND usage > 0").to_a
     end
 
-    def new_request
```
majewsky: You deleted the `new_request` action. Please add it back.

plugins/resource_management/app/controllers/resource_management/project_resources_controller.rb
```diff
@@ -21,7 +23,59 @@ def index
...
+                  @project_resource.current_quota == @project_resource.data_type.parse(value)
+              @project_resource.add_validation_error(:current_quota, "wrong value: because the reduced quota value is the same as your current quota")
+            else
+              @project_resource.approved_quota = @project_resource.data_type.parse(value)
+              @project_resource.current_quota = @project_resource.data_type.parse(value)
+            end
+          rescue ArgumentError => e
+            @project_resource.add_validation_error(:current_quota, 'is invalid: ' + e.message)
+          end
+        end
```
majewsky: This paragraph would be easier to read if you put `@project_resource.data_type.parse(value)` into a variable at the start. (This also applies to `domain_admin#reduce_quota`.)

plugins/resource_management/app/controllers/resource_management/domain_admin_controller.rb
majewsky: Nevermind. I was just confused because the `else` has an explicit `render action: 'reduce_quota'`, which is the same as what's happening here implicitly.

plugins/resource_management/app/controllers/resource_management/domain_admin_controller.rb
```diff
@@ -68,6 +70,55 @@ def update
       end
     end
 
+    def reduce_quota
+      # Note: this function is used in two ways
+      # first show the reduce quota window
+      # second: reduce the quota
```
majewsky: Ah, that's the source for my other confusion down below. Can you split that into two actions, please?

plugins/resource_management/config/policy.json
```diff
@@ -33,6 +35,8 @@
   "resource_management:project_resource_create_request": "rule:access_to_project",
   "resource_management:project_resource_new_package_request": "rule:access_to_project",
   "resource_management:project_resource_create_package_request": "rule:access_to_project",
+  "resource_management:project_resource_reduce_quota": "rule:access_to_project",
+  "resource_management:project_resource_confirm_reduce_quota": "rule:access_to_project",
```
majewsky: I think this should be `rule:project_admin`. Members should not be able to reduce their project's quota. On the other hand, it is kind of understandable that you chose this in symmetry to "request quota", which is open to everyone. Since we're going to be billing on quota, we should restrict all quota-changing actions to `rule:project_admin`.

PR #57, opened 2017-03-01
add no delimiter option to format function

hgw77:


spec/lib/core/data_type_spec.rb
```diff
@@ -20,6 +20,10 @@
       expect(datatype_number.format(12345)).to     match(/^12[[:space:]]345$/)
       expect(datatype_number.format(1234567)).to   match(/^1[[:space:]]234[[:space:]]567$/)
       expect(datatype_number.format(123456789)).to match(/^123[[:space:]]456[[:space:]]789$/)
+      # check with no delimiter (used for editable fields)
+      expect(datatype_number.format(12345, false)).to     match(/^12345$/)
+      expect(datatype_number.format(1234567, false)).to   match(/^1234567$/)
+      expect(datatype_number.format(123456789, false)).to match(/^123456789$/)
```
majewsky: Can you make the second argument an `options` hash instead? `format(123, false)` is less readable than e.g. `format(123, delimiter: false)`.

PR #59, opened 2017-03-03
[core] extend base model with pretty time formatters

Carthaca:
and make use of them

@andypf : why can't I use this at https://github.com/sapcc/elektra/blob/master/plugins/shared_filesystem_storage/app/assets/javascripts/shared_filesystem_storage/components/shares/show.coffee#L62 and similar at share_snapshots?

maybe https://github.com/sapcc/elektra/blob/master/plugins/object_storage/app/views/object_storage/objects/show.html.haml#L23 is another candidate

### DIFF
--- a/lib/core/service_layer/model.rb
+++ b/lib/core/service_layer/model.rb
@@ -157,12 +157,20 @@ module Core
 
       def created_at
         value = read("created") || read("created_at")
-        Time.parse(value) if value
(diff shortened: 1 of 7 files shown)


andypf:
show.coffee is javascript!

Carthaca:
thx, so that js can be tackled separately

and this helper here is not needed at obj storage

PR #63, opened 2017-03-07
WIP: handle backend failures better when trying to apply quotas

majewsky:
Please do not merge yet, I want to get rid of that `<p class="alert alert-error" id="error_message">` first.


plugins/resource_management/app/controllers/resource_management/domain_admin_controller.rb
```diff
@@ -226,9 +227,10 @@ def approve_package_request
         res.save!
       end
 
-      services.resource_management.apply_current_quota(@project_resources)
+      @services_with_error = services.resource_management.apply_current_quota(@project_resources)
       services.inquiry.set_inquiry_state(@inquiry.id, :approved, 'Approved')
-      render action: 'approve_request'
+      
+      render action: 'approved_request'
```
majewsky: The view having a different name than the controller is needlessly confusing, esp. for a single-letter difference.

plugins/resource_management/app/views/resource_management/domain_admin/approved_request.js.erb
```diff
@@ -0,0 +1,11 @@
+<% if @services_with_error.empty? %>
+  $('#set_quota_error').addClass('hide');
+<% else %>
+  $(".flashes").html("<%= escape_javascript(render "set_quota_error") %>");
```
majewsky: Please `.append()` instead of overwriting, and don't hide existing errors.

PR #64, opened 2017-03-07
Empty flashes

hgw77:
on click flashes are hidden, good or bad idea? ;-)


edda:
As just discussed, will close this without merging. Arturo is currently working on a general solution for hiding flashes. Click-to hide is not recommendable because we sometimes have links in flashes that we want to be able to click without hiding the flash, that is to say, hiding should be done with a dedicated hide button/link.

PR #65, opened 2017-03-09
[resmgmt] remove confusing dns records resource

Carthaca:
Hi @majewsky, can you have a look and clean up the db afterwards please?


majewsky:
Looks good. Will merge tomorrow. (If not, please poke me.)

Do we need to do anything to reset the records quotas that we already wrote into Designate?

Carthaca:
Good point, to avoid surprises setting the records-quota to our `20 * recordsets-quota` everywhere would be nice

majewsky:
Ah, but that should be covered by a simple re-apply of the existing recordset quotas with the new driver. So no extra code needed.

Carthaca:
Yes, you are planning to do that re-apply anyhow, right?

majewsky:
I've added code to perform the reapply of the recordset quotas. (This cannot be done as a rake task or similar because I need the `services` object which is only accessible within controllers.)

PR #68, opened 2017-03-13
added kubernetes to the powered by section.

urfuwo:


urfuwo:
added Kubernetes to the powered by jingle

PR #69, opened 2017-03-13
[WIP] adjust resmgmt policy.json for new resource_admin/viewer roles

majewsky:
Do not merge unless @ruvr has rolled out the respective roles.

@ruvr A second set of eyes would be nice to verify if I got the policy rules right.


ruvr:
lgtm, although I don't know about the domain_id:nil stuff.. 
that' something for AndreasP

the new roles & permissions have been seeded

majewsky:
@andypf Care to double-check?

PR #76, opened 2017-03-16
make domain quotas binding

majewsky:
Whenever a domain admin adjusts project quotas, ensure that they cannot exceed their domain quota allocation.


hgw77:
lgtm 👍

PR #81, opened 2017-03-22
Resource management/sort details

hgw77:


majewsky:
@hgw77 Very nice. Can you also add the sorting buttons to the cloud admin's "Details" view?

hgw77:
oh yes I can do that :-)

PR #93, opened 2017-05-10
async emptying for containers

hgw77:
### emptying in process
empty-container1

### done
empty-container2

### note for eventual consistency 
empty-container3


plugins/object_storage/config/routes.rb
```diff
@@ -7,7 +7,8 @@
     member do
       get  :confirm_deletion
       get  :confirm_emptying
-      post :empty
+      post :pre_empty
+      get  :empty
```
majewsky: Should also be `post` since it models an action rather than a resource.

PR #100, opened 2017-05-18
[cinder] allow to reset volume attach_status + hint

Carthaca:
@andypf : Sorry! I knew this would come back to me - turned out to be very soon :)


Carthaca:
@edda any suggestions for making this info prettier or okay with you? https://github.com/sapcc/elektra/blob/eef5849b36c52610c36cadcc5502699b8bf949f8/plugins/block_storage/app/views/block_storage/volumes/new_status.html.haml#L3

edda:
@Carthaca I turned the info help link into an alert. I think this warrants it as people can potentially break things here.

PR #107, opened 2017-06-09
[compute] add instance action tab to show for admins

Carthaca:
admins get here from looking up directly, so can't trigger actions
at the list

e.g. to delete stuck instances found via object lookup server

@edda : what do you think? is this pretty enough?


edda:
@Carthaca What do I need to do to have a look at this? Do I need a special role?

Carthaca:
you need `cloud_compute_admin`

edda:
Okay, got it :) I changed it so the action is rendered as a button. We can revisit if at some point we have so many actions in there that buttons don't make sense anymore. But with only the terminate action I'd do it like this...

Carthaca:
thanks

PR #110, opened 2017-06-14
[resource_management] bugfix in domain_admin_controller

andypf:
@majewsky Please take a look. We discovered an error, if a service is not available then @domain.find_resource returns nil and it bangs.


hgw77:
@majewsky: for your background information, the problem happened when a resource was deactivated in limes

majewsky:
Not sure what you mean by "deactivated", but lgtm (or rather, "looks inoffensive to me").

hgw77:
there was no data for Shared Filesystem Storage available and so the resource collection was disabled

majewsky:
Hm, in that case you should actually remove the ResourceConfig entries from `lib/resource_config.rb`, or at least comment them out.

PR #145, opened 2017-10-10
prevent change of classname from plural to singular

hgw77:
with that changes I can create a maia or masterdata plugin without problems :-)


andypf:
lgtm

PR #163, opened 2017-11-21
[compute] list instance actions in policy and allow for member

Carthaca:
Hi @andypf, can you please have a look?


andypf:
lgtm!

PR #216, opened 2018-02-01
Compute::Flavor.save: do not fail when creating new flavor

majewsky:
This is not working yet, but no idea why. There seems to be something horribly broken in the exception handling. For example, if I do this instead:

```ruby
begin
  @service.delete_flavor(id) unless id.blank?
rescue => e
  pp(e.class)
  pp(e.code)
  raise StandardError, 'manual error'
end
```

I see the output of the `pp()`, but the UI shows the error message from `e` rather than "manual error". WTF.


andypf:
lgtm

PR #226, opened 2018-02-21
remove the technical user workaround, since it has been replaced by a…

ruvr:
… CAM self-service

Not sure if we should remove it (as in this pull request) or initially just disable it via a policy change..


andypf:
lgtm. I prefer to remove the obsolete code. So merge pull request!

ruvr:
can we please merge this change? creating technical users isn't supported by our keystone domain configuration anymore.

PR #243, opened 2018-03-28
discourage remote security groups

majewsky:
grafik

I previously considered a design where the user has to confirm that they read the documentation by checking a checkbox (like when agreeing to ToS), but that is probably too drastic.


majewsky:
Thanks. Like with #244, I'm not merging this yet. I want to check with Falk that our docu is in a good state first.

PR #263, opened 2018-04-12
Don't overwrite already exported password in openrc generated file

kayrus:
Especially it is useful when using [direnv](https://github.com/direnv/direnv)


edda:
This is a good idea. I see two potential problems with this though: 

1. When you change your password you need to know that you have to delete or update the env variable with the new password since you won't be asked for it anymore.

Idea: Maybe add an else case that prints a message like "Using previous OS password. If you have changed your password please delete or update the env variable 'OS_PASSWORD'"

2. Similar but harder to fix: if a person uses multiple user accounts then the password most likely wouldn't be the same. So in that case they'd want to be asked for the password every time. Not sure how we would fix this in a way that is convenient for the user...

So I fear that this change would result in more support tickets for us...

Issue #201, opened 2018-01-25
Admin Service: Find projects by floating IP or DNS record


edda:
Implement a service for admins that allows an input of floating IP or DNS record. The service should then find which project the input belongs to and display the following information about the project:

- Project domain
- Project name
- Project ID
- List of users with the role admin

Additionally retrieve some information about the object the floating IP or DNS record is attached to:

- Type of object (server, load balancer, ...)
- Name
- Backend information (e.g. building block or vcenter info, ...)

The service shall be available (via link) from the domain start page and only visible/useable by users with the new security_ops role (already available in staging).

There is already a branch ([reverse-lookup](/sapcc/elektra/tree/reverse-lookup)) where I did some initial work before deciding that this was too cumbersome to do in rails and that it needed to be a react app. The code is in the lookup plugin under `reverse_lookup` (controller and views)

For the user lists my idea was to get users with the role and then separately get groups with the role and make the groups expandable (so the users in the group are only loaded if the group is expanded).

PR #265, opened 2018-04-16
Admin Service: Find projects by floating IP or DNS record

ArtieReus:
Service for admins that allows an input of floating IP or DNS record. The service should then find which project the input belongs to and display the following information about the project:

Project domain
Project name
Project ID
List of users with the role admin

Additionally retrieve some information about the object the floating IP or DNS record is attached to:

Type of object (server, load balancer, ...)
Name
Backend information (e.g. building block or vcenter info, ...)
The service shall be available (via link) from the domain start page and only visible/useable by users with the new security_ops role (already available in staging).


ArtieReus:
Fixes #201

PR #298, opened 2018-06-06
Removed Instance Snapshot Warning

sapliedtke:
Problem is fixed, see jira ticket https://issues.wdf.sap.corp/browse/CCM-938


edda:
Thank you! Closing this, as I had already removed the warning but not pushed yet.

PR #335, opened 2018-08-03
Fix automation_script permissions typo

kayrus:
Customer with only one `member` role was not allowed to create an automation for a node.

```
{"error":"You are not authorized to view this page. Please check (role assignments) if you have one of the           following roles: admin."}
```


edda:
lgtm

PR #350, opened 2018-08-24
allow role-assignment-management of baremetal & cfm roles

ruvr:
Can you please check if there is any reason for not allowing management of these roles in elektra?
The S4 team is stuck, since they are missing these..


edda:
Checked and learned that these roles are not to be whitelisted. Will close this PR :)

PR #356, opened 2018-08-30
Reduce quota adjustments

hgw77:
capture

### DIFF
--- a/plugins/resource_management/app/controllers/resource_management/project_resources_controller.rb
+++ b/plugins/resource_management/app/controllers/resource_management/project_resources_controller.rb
@@ -42,6 +42,8 @@ module ResourceManagement
     end
 
     def confirm_reduce_quota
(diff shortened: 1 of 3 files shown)


plugins/resource_management/app/controllers/resource_management/project_resources_controller.rb
```diff
@@ -42,6 +42,8 @@ def show_area(area = nil)
     end
 
     def confirm_reduce_quota
+      puts "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
+      pp params
```
majewsky: I believe this can go. :)

hgw77: ups! thanks

and yes it is a requirement from @reimannf to reduce a quota to the usage with less steps

Issue #330, opened 2018-07-23
Swift: container creation with non-standard chars leads to inaccessible container


edda:
It seems that there are no restrictions as to which characters can be used to name a swift container. However using special chars ( probably related to being URL compatible or not ) leads to the container not being accessible in Elektra via URL.

To reproduce:
Create a new swift container via Elektra or CLI with the name: `echo 1; rm -rf *)` 
** Result:** 
Container is created but not accessible via Elektra - error message when trying to access is - 
```
Object not Found
The object you are looking for doesn't exist. Please verify your input. ()
```

I think this needs some investigation as to what exactly the problem is and whether swift really does not impose any sort of restrictions on naming containers. Then see how we can solve the display issue.

edda:
Additional info from reimannf:

There are the following restrictions on container names (see https://docs.openstack.org/swift/latest/api/object_api_v1_overview.html):Length of container names256 bytesCannot contain the / character. | Length of container names | 256 bytes | Cannot contain the / character.

`Length of container names | 256 bytes | Cannot contain the / character.`

Before restricting the allowed characters, we should check whether the elektra plugin correctly UTF-8 encodes/decodes container names.
E.g. sample output of the python swift client, see the URL in the curl:

```
swift stat --debug 'echo 1; rm -rf *)'
...
DEBUG:swiftclient:REQ: curl -i https://objectstore-3.staging.cloud.sap:443/v1/AUTH_xxx/echo%201%3B%20rm%20-rf%20%2A%29 -I -H "X-Auth-Token: TOKEN"
DEBUG:swiftclient:RESP STATUS: 204 No Content
DEBUG:swiftclient:RESP HEADERS: {u'Content-Length': u'0', u'X-Container-Object-Count': u'0', u'Date': u'Mon, 23 Jul 2018 10:41:16 GMT', u'Accept-Ranges': u'bytes', u'Server': u'nginx/1.10.2', u'Last-Modified': u'Mon, 23 Jul 2018 10:40:52 GMT', u'Connection': u'keep-alive', u'X-Timestamp': u'1532342451.88471', u'X-Trans-Id': u'tx8592507f3bd6429cb56cc-005b55b0cc', u'X-Container-Bytes-Used': u'0', u'X-Storage-Policy': u'default', u'Content-Type': u'application/json; charset=utf-8', u'X-Openstack-Request-Id': u'tx8592507f3bd6429cb56cc-005b55b0cc'}
               Account: AUTH_xxx
[...]

PR #360, opened 2018-09-05
Object storage and weird names

hgw77:
to support in container/folder/file any kinds of chars (except `/`)


hgw77:
realted to #330

PR #422, opened 2019-01-10
added generic remote console support

notandy:
This PR will implement the new remote console api:
https://developer.openstack.org/api-ref/compute/?expanded=show-console-output-os-getconsoleoutput-action-detail,get-vnc-console-os-getvncconsole-action-deprecated-detail#create-remote-console

and also unify shellinabox/vnc console to generic "remote_console" function.
Also enabled MKS by default, please wait a week before rollout to ensure mks is everywhere enabled.


notandy:
The MKS web-console is now deployed in all reagions, this PR is ready to merge

PR #426, opened 2019-01-17
Filter params by regex expresion

ArtieReus:
In that way is posible to match any kind of composed params names with the ones defined on the list.


ArtieReus:
@andypf lgty? This is how I did it in automation.

andypf:
lgtm!

PR #443, opened 2019-02-20
Resource management fixes

hgw77:


hgw77:
if user has only viewer rights do not show settings button and adjust bar size

PR #444, opened 2019-02-22
collect pagination data once with 0,5 sec delay when checking completed

ArtieReus:
The pagination fetcher triggered by the search field had following issue "Maximum call stack size exceeded" coming from the application.js file. This is due that the loadAllOnSearch is being called continuously until end of stack. With a delay and triggering the fetcher once fix the problem.


ArtieReus:
@andypf lgty?

andypf:
👍

PR #453, opened 2019-03-19
new resource-management plugin (part 1: customer workflows)

majewsky:
I'm leaving the approver workflows out of this for now, so that we can get the first part in.

There are also some "TODO" comments in the code that I have yet to address, hence the draft status. Most prominently, three big things are missing:
- [x] creation of inquiry items in Elektra upon "Submit" of "Edit" dialog
- [x] for shared resources: cluster list in "Details" view on cluster level
- [x] integration with the rest of Elektra (esp. main menu should link to the new plugin)


majewsky:
This is ready for review.

@andypf @edda Please have a look at the changes outside of `plugins/resources/`.

@hgw77 I'll send you an invite for a testing session.

PR #464, opened 2019-04-15
new wizard step: Initialize Project Resources

majewsky:
The new implementation only requests quotas within auto-approval limits, so no request workflow is needed.

grafik

The screenshot shows an example situation where the domain has run out of Block Storage quota, so the respective package cannot be selected by the user. (The tooltip is shown because I hovered over the Block Storage package while taking the screenshot.)

### DIFF
--- a/plugins/resources/app/assets/stylesheets/resources/_application.scss
+++ b/plugins/resources/app/assets/stylesheets/resources/_application.scss
@@ -218,11 +218,6 @@ div.edit-quota-input {
       font-size: 18px;
       line-height: 18px;
       font-weight: bold;
-
-      i.fa {
-        /* different icons apparently have inconsistent widths */
-        width: 16px;
-      }
     }
 
     .muted-resources {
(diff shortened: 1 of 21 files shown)


majewsky:
@reimannf @hgw77 Rebased to leave the old package request workflow intact, as discussed.

hgw77:
👍

plugins/resources/app/assets/stylesheets/resources/_application.scss
```diff
@@ -169,3 +169,70 @@ div.edit-quota-input {
...
+    }
+
+    h3 {
+      margin: 0 0 14px 0;
+      font-size: 18px;
+      line-height: 18px;
+      font-weight: bold;
+
+      i.fa {
+        /* different icons apparently have inconsistent widths */
```
edda: There's a css class you can add to icons to make them all the same width: `fa-fw`

PR #477, opened 2019-05-29
added support for role_admin and role_viewer

ruvr:
Could you please check if this does the job or if more changes are needed?
This is related to https://issues.wdf.sap.corp/browse/CCM-9001


ruvr:
Could you please check & merge if OK? S4 needs this sort of urgently...

PR #487, opened 2019-07-25
Policymod for BSO - compute_admin without securitygroup admin

rajivmucheli:
Hi @edda 

Please validate and approve this PR.

Regards,
Rajiv

### DIFF
--- a/plugins/compute/config/policy.json
+++ b/plugins/compute/config/policy.json
@@ -4,7 +4,7 @@
   "compute_viewer": "role:compute_viewer",
   "compute_admin": "role:compute_admin",
   "compute_admin_wsg": "role:compute_admin_wsg",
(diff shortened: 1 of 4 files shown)


plugins/compute/config/policy.json
```diff
@@ -3,6 +3,7 @@
   "member": "role:member or role:Member",
   "compute_viewer": "role:compute_viewer",
   "compute_admin": "role:compute_admin",
+  "compute_admin_wsg": "role:compute_admin_wsg",
   "context_is_compute_admin": "rule:context_is_cloud_compute_admin or rule:compute_admin",
```
edda: I understood that this role is supposed to be a compute_admin but without the permission to edit security groups. So you will have to make sure in this line here to add `or rule:compute_admin_wsg` to the `context_is_compute_admin` rule. Otherwise they won't be a compute admin.

PR #490, opened 2019-08-01
[BSO Policy] network_viewer should list instance

rajivmucheli:
Hi @edda,

BSO requests network_viewer + securitygroup_admin should allow user to list servers to edit security groups. Hope this works?

Regards,
Rajiv


edda:
Closed because it doesn't make sense to make this change when the compute_viewer role exists and gives exactly those permissions. Also we should be very careful with introducing role checks using roles from another Openstack service.

PR #493, opened 2019-08-29
redirect /:domain_id/:project_id -> /:domain_id/:project_id/home

databus23:
There were some concerns that adding this redirect would break other routes but in my local tests it works and I couldn't find any broken project plugin links.


databus23:
@andypf could you take a look and validate that this works without breaking stuff?

PR #624, opened 2020-03-11
adds first version of metrics=> grafana as a service

stefanhipfel:


plugins/metrics/app/controllers/metrics/application_controller.rb
```diff
@@ -12,5 +12,41 @@ def index
...
+      request = Net::HTTP::Post.new(uri.request_uri, header)
+      request.body = grafana.to_json
+    
+      # Send the request
+      response = https.request(request)
+      if response.code == '200' or response.code == '201'
+        body = JSON[response.body]
+        print body['hostname']
+        redirect_to "https://"+ body['hostname']
+      end
```
stefanhipfel: todo: error handling!

plugins/tools/config/policy.json
```diff
@@ -6,5 +6,6 @@
   "tools:universal_search_role_assignments": "rule:cloud_admin or role:cloud_support_tools_viewer",
   "tools:universal_search_user_role_assignments": "rule:cloud_admin or role:cloud_support_tools_viewer or user_id:%(user.id)s",
   "tools:switch_to_project": "not project_id:%(project.id)s",
-  "tools:show_scope_object": "rule:cloud_admin or role:cloud_support_tools_viewer"
+  "tools:show_scope_object": "rule:cloud_admin or role:cloud_support_tools_viewer",
+  "metrics:application_gaas": "role:cloud_support_tools_viewer"
```
edda: Please add this line into the policy file of the metrics plugin, not here. The metrics policy file is here: https://github.com/sapcc/elektra/blob/master/plugins/metrics/config/policy.json

plugins/metrics/app/controllers/metrics/application_controller.rb
stefanhipfel: I am doing this now server side. It will return the url once the Grafana pod/ingress is up and running.

edda: Yes, sorry, I wrote this comment a while ago for an older version of the pull request. It's outdated now. I'm not sure why i posted this again with the review. Maybe I forgot to finish the review last time and my comment never showed up. My actual comment this time is the one about the policy entry :)

plugins/metrics/app/controllers/metrics/application_controller.rb
```diff
@@ -1,16 +1,118 @@
 # frozen_string_literal: true
-
 module Metrics
   class ApplicationController < DashboardController
+    helper_method :grafana_available
     authorization_context 'metrics'
     authorization_required
-
+    extend ErrorRenderer
```
hgw77: do we need that? I removed it and tested the error case and the error was rendered correctly

plugins/metrics/app/views/metrics/application/index.html.haml
```diff
@@ -28,8 +27,18 @@
...
+      %p
+        Grafana is a system to display Maia metrics. It provides a pre-configured set of dashboards. [ Currently only available in qa-de-1 and eu-ru-1 ].
 %p
-  = link_to :maia, class: "btn btn-primary", target:'_blank' do
+
+  = link_to :maia, class: 'btn btn-primary', target:'_blank' do
     %i.fa.fa-external-link
     Open Maia Dashboard
+  - if grafana_available && current_user && current_user.is_allowed?('metrics:application_gaas')
+    = link_to :gaas, class: 'btn btn-primary', style: 'margin-left: 5px', title: 'only available for qa-de-1 and eu-ru-1', target:'_blank' do
```
hgw77: For the first time click, the user will see an empty new tab and needs to wait until grafana is done. We have the same if an error happened. Than we will see the error page in the new tab. 

Better would be to have an ajax call that is doing the error handling and polling. Thats not a must have, the current version is working well but maybe a good improvement for later ;-)

plugins/metrics/app/controllers/metrics/application_controller.rb
stefanhipfel: removed it. not sure if needed or not

PR #648, opened 2020-05-11
[Glance][Policy] Add rule:image_admin, remove duplicate policy proper…

rajivmucheli:
…ty image:image_owner

Hi @edda 

Please validate and approve.

Regards,
Rajiv

### DIFF
--- a/plugins/image/config/policy.json
+++ b/plugins/image/config/policy.json
@@ -9,8 +9,8 @@
   "image:image_is_shared": "'shared'==%(image.visibility)s",
   "image:image_is_community": "'community'==%(image.visibility)s",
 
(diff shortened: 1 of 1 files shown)


plugins/image/config/policy.json
```diff
@@ -1,19 +1,19 @@
 {
   "image_viewer": "role:image_viewer",
-  
-  "image:image_owner": "project_id:%(image.owner)s",
```
edda, line 4: Are you sure you want to remove the image owner rule? That means that anybody with image_admin can delete any image even ones they haven't created. Also if you do remove the rule you need to remove it from the rules further down as well. Currently it's still being used e.g. in the image_delete rule.

Carthaca:
The changes are already in master now

PR #713, opened 2020-08-28
[lbaas2]: Quota usage info and help info with links reactivated

ArtieReus:


ArtieReus:
@notandy would you check if the info text it is actual? Just copied from the old plugin. Do you have more links to show in the info since at least one of them is not working anymore.

ArtieReus:
@notandy  Sorry... I have to merge. We should check that later. I will open a ticket for Info text and links review.

PR #732, opened 2020-09-07
[shared_filesystem] show cloud admin pools for all types

Carthaca:
we will get more than one type, also in production soon and want to have an overview which pool is related to what type


andypf:
lgtm!

PR #781, opened 2020-11-02
Fixed main modal view scroll when comming from a infoDialog

ArtieReus:
Closing the confirm dialog when the main modal view still open removes the class "modal-open" from body which prevent the main modal view to be scrolled again


andypf:
LGTM

PR #782, opened 2020-11-06
[cronus]: add cronus roles

kayrus:
cc @majewsky


majewsky:
@andypf Can we safely merge this? If this shows up in the UI regardless of whether the roles actually exist, we shouldn't merge yet because Cronus is not deployed to prod yet. If the UI is smart enough to skip roles that don't actually exist in Keystone, this would be safe to merge.

andypf:
LGTM

PR #820, opened 2021-01-06
[Barbican][Policy] Add support for yaml format as json format support…

rajivmucheli:
… is depcreated in Victoria release

Hi @edda,

Openstack Victoria release has deprecated policy.json support, hence we need to switch policy.yaml.

Reference : https://docs.openstack.org/releasenotes/oslo.policy/victoria.html

I have switched openstack barbican to support policy.yaml in qa-de-1, it works well with CLI but i get permission denied if interacted via GUI. Could you please enable this fix only in qa-de-1 to test before rolling it to production ? Could you suggest how do we coordinate the rollout, as i plan to upgrade barbican 4 regions per week starting next week.

Regards,
Rajiv

### DIFF
--- a/plugins/key_manager/config/policy.yaml
+++ b/plugins/key_manager/config/policy.yaml
@@ -1,0 +1,21 @@
+# WARNING: Below rules are either deprecated rules
+# or extra rules in policy file, it is strongly
+# recommended to switch to new rules.
+"context_is_cloud_keymanager_admin": "role:cloud_keymanager_admin"
+"member": "role:member or role:Member"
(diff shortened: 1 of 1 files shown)


rajivmucheli:
i guess barbican will need a custom path here ,   paths = ["config/policy.yaml"]

https://github.com/sapcc/elektra/blob/master/config/initializers/monsoon_openstack_auth.rb#L2-L8

edda:
Hi @rajivmucheli, 

These changes in OpenStack are not relevant for Elektra. We have our own policy engine here which evaluates the policy files. There's no need to change it to YAML on Elektra side.

PR #832, opened 2021-02-22
Lbaas basic UI retirement

ArtieReus:


plugins/resource_management/app/controllers/resource_management/automation_controller.rb
```diff
@@ -18,7 +18,7 @@ def dump_data
       legacy_names = {
         :compute        => 'compute',
         :dns            => 'dns',
-        :loadbalancing  => 'loadbalancing',
+        :loadbalancing  => 'lbaas2',
```
ArtieReus, line 21: @majewsky, @hgw77  would you check please if this change is correct.

plugins/resources/app/controllers/resources/quota_usage_controller.rb
```diff
@@ -8,7 +8,6 @@ class QuotaUsageController < ::Resources::ApplicationController
     QUOTA_RESOURCES = {
       compute: {service_type: :compute, resource_names: [:instances, :cores, :ram]},
       block_storage: {service_type: :volumev2, resource_names: [:volumes, :snapshots, :capacity]},
-      loadbalancing: {service_type: :network, resource_names: [:loadbalancers, :listeners, :pools, :l7policies]},
```
ArtieReus, line 11: @majewsky, @hgw77 have any side effects removing the old plugin and introducing the new one? The name of the plugin has changed.

plugins/lbaas2/app/javascript/app/App.js
```diff
@@ -10,8 +10,8 @@ const App = () => {
     <React.Fragment>
       <div className="bs-callout bs-callout-info bs-callout-emphasize">
         <p>
-          This is the new beta UI based on Octavia driver for Neutron LBaaS.
-          Feel free to try it! We are happy to get your feedback on our slack
+          This is the new UI based on Octavia driver for Neutron LBaaS.
```
ArtieReus, line 13: @edda does it make sense to still keep an info box at the top of the plugin or we can get rid of it?

plugins/resources/app/controllers/resources/quota_usage_controller.rb
majewsky: If I remember correctly, this is @andypf's code. But I don't know what side effect there should be.

plugins/resource_management/app/controllers/resource_management/automation_controller.rb
ArtieReus: @majewsky please check this part if it makes any side effects. Thx

majewsky: This controller has not been in use since 2017, so no.

PR #833, opened 2021-02-24
fix error controller when comming from js request

ArtieReus:


app/controllers/errors_controller.rb
```diff
@@ -57,7 +66,7 @@ def details
         h[:description] = i18n.t("#{key}.description", default: description).html_safe
       end
 
-      h[:env]       = request.env
+      # h[:env]       = request.env # commented since it blows up the json redering. If needed we should select needed attributes from the env
```
edda, line 69: And taking it out doesn't break anything else?

ArtieReus: nop. This is just extra information we print out when rendering the error.

andypf: lgtm

andypf (approved):

PR #877, opened 2021-04-21
Lbaas2 healthmonitor IPs + improvements

ArtieReus:
- Display of allowed adresses pair in the healthmonitor section
- Corrected Plugin name
- JSONView with memorize
- Fetch jus the loadbalancer when vip_port_id not send as param


notandy:
👍

PR #920, opened 2021-07-16
[Barbican] Set Expiration field as mandatory

rajivmucheli:


rajivmucheli:
Based on `If set, the secret will not be available after this time` how is this calculated ? would it impact existing certificates ?

rabindra-chauhan (approved):
approved

PR #931, opened 2021-08-31
Email service

sirajudheenam:


andypf:
I can't check all 72 files for correctness, but it's a plugin. If there are problems we can deactivate this plugin relatively quickly. We do this by using the available? from the service layer to decide whether the link to the plugin is displayed.

However, I don't see a corresponding entry in /config/navigations/services_navigation.rb

How or where can the user see the UI of this plugin?

sirajudheenam:
Navigation is at https://github.com/sirajudheenam/elektra/blob/master/config/navigations/services_navigation.rb#L285-L295
UI can be seen in this URL: /monsoon3/cc-demo/email-service/emails. It is meant to be hidden until tests are completed.

ArtieReus (changes requested):
It seems that there is a syntax error `Unexpected token: keyword (const)` when starting the app from your branch and first accessing the landing page. If I remove this file `plugins/email_service/app/assets/javascripts/email_service/email.js` and I restart the app the landing page can be compiled correctly. 
Would you please first fix that before we continue with the review. Thx!

ArtieReus (changes requested):
Hi @sirajudheenam! Please just keep js files used for the plugin. Remove, no comment, not used files. Thx!

ArtieReus:
Hi @sirajudheenam! Is the JS not needed for the plugin to work properly? If so, then please instead of commenting out the code fix it so that I can test the plugin as designed.

ArtieReus (changes requested):
Hi @sirajudheenam! Please see following example to build an ui switcher based on a role:

1. Move the application index action to new controller called ex. `web_controller`
2. Move the index view to a new folder called ex. `web` under views
3. Change the routes so the entry point stays as it was before. Ex: `get '/' => 'web#index', as: :index `
4. Add following methods to restrict access and switch ui in the application controller
 ```
   def ui_switcher
      if current_user.has_role?('cloud_support_tools_viewer')
        redirect_to emails_path
      end
    end

    def restrict_access
      unless current_user.has_role?('cloud_support_tools_viewer')
        redirect_to index_path
      end
    end 
```
5. Use the before_action :restrict_access to redirect users without this role to old ui
6. Use this before_action :ui_switcher to switch users with this role to the new ui
7. Write some tests :)

sirajudheenam:
> Hi @sirajudheenam! Please see following example to build an ui switcher based on a role:
> 
>     1. Move the application index action to new controller called ex. `web_controller`
> 
>     2. Move the index view to a new folder called ex. `web` under views
> 
>     3. Change the routes so the entry point stays as it was before. Ex: `get '/' => 'web#index', as: :index `
> 
>     4. Add following methods to restrict access and switch ui in the application controller
> 
> 
> ```
>   def ui_switcher
>      if current_user.has_role?('cloud_support_tools_viewer')
>        redirect_to emails_path
>      end
>    end
> 
>    def restrict_access
>      unless current_user.has_role?('cloud_support_tools_viewer')
>        redirect_to index_path
>      end
>    end 
> ```
> 
>     1. Use the before_action :restrict_access to redirect users without this role to old ui
> 
>     2. Use this before_action :ui_switcher to switch users with this role to the new ui
> 
>     3. Write some tests :)

1,2 and 3 steps are completed.

PR #991, opened 2021-11-19
Rename VNC console to Web Console

fwiesel:
Originally, we were using a VNC session, but that is switched to an
MKS one since some time now.
And for ironic, we do use some different service altogether.
So we better name everything technology agnostic in order to avoid
confusion


fwiesel:
I am not sure about the name "Web" console, but I think the "VNC" part needs to go.
Some feedback on the naming is appreciated.
I haven't run it yet, so it may not even work

edda (approved):
lgtm

PR #1003, opened 2021-11-26
[resources] Autoscaling fix

### LINKED ISSUE #992: quota autoscaling: UI input values cross-bleed between projects
To reproduce:

1. In a domain where quota autoscaling is enabled, go to Resource Management > Autoscaling > Config by Project.
2. For two different projects, set up autoscaling for the same resource with the same configuration values, including a minimum-free value (e.g. "1 % but at least 2 units").
3. Go to the config for the first project, edit the minimum-free value (e.g. "1% but at least 3 units"), save and close the project config.
4. Go to the config for the second project.

In the last step, the config for the second project should be shown, including the unchanged minimum-free value that was set in step 2. **However,** I can actually see the value that was changed in the other project in step 3.

The UI still seems to have the right understanding of what the actual config values are, since the UI correctly chooses whether to display "Save" and "Cancel" buttons, respectively. But the config view as well as the editor starts out with the wrong values bleeding over from the previously-edited project config.

ArtieReus:
Long story short...
tempMinfree was created once when the modal view was loaded. 
https://github.com/sapcc/elektra/blob/master/plugins/resources/app/javascript/app/components/autoscaling/config_ng/subscope_item.js#L26

When switching the different autoscaling projects react.useffect was triggered updating the minFree in the State with the NOT updated tempMinFree set when the modal view was opened for first time. The trigger was not the tempMinFree itself but the unit.
https://github.com/sapcc/elektra/blob/master/plugins/resources/app/javascript/app/components/autoscaling/config_ng/subscope_item.js#L48

I didn't see any benefit of using this tempMinFree so I removed it and instead the state is being updated directly. @andypf can do a double check before merging.

cc @majewsky 
fixes #992

### DIFF
--- a/plugins/resources/app/javascript/app/components/autoscaling/config_ng/subscopes.js
+++ b/plugins/resources/app/javascript/app/components/autoscaling/config_ng/subscopes.js
@@ -1,8 +1,8 @@
-import { Table, FormControl, Button, Row, Col } from "react-bootstrap"
-import SubscopeItem from "./subscope_item"
-import { t } from "../../../utils"
-import { parseConfig, generateConfig, isUnset } from "../helper"
-import { Unit } from "../../../unit"
+import { Table, FormControl, Button, Row, Col } from "react-bootstrap";
+import SubscopeItem from "./subscope_item";
+import { t } from "../../../utils";
+import { parseConfig, generateConfig, isUnset } from "../helper";
+import { Unit } from "../../../unit";
 
 const styles = {
   detailsWrapper: {
@@ -22,7 +22,7 @@ const styles = {
     display: "flex",
     justifyContent: "space-between",
   },
-}
+};
 
 /****************************************
  * Details View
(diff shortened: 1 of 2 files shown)


majewsky:
Can you please commit the style changes (the added semicolons) to master right away? I would like to review the diff, but right now it's gigantic.

plugins/resources/app/javascript/app/components/autoscaling/config_ng/subscopes.js
```diff
@@ -276,10 +276,11 @@ const AutoscalingConfigNgSubscopes = ({
...
       a.service < b.service ? -1 : a.service > b.service ? 1 : 0
     );
-  }, [subscopesState, searchTerm]);
+  }, [JSON.stringify(subscopesState), searchTerm]);
 
   const editModeCount = React.useMemo(
-    () => filteredSubscopes.filter((s) => s.editMode).length
+    () => filteredSubscopes.filter((s) => s.editMode).length,
+    [JSON.stringify(filteredSubscopes)]
   );
```
majewsky: Why is this code (both the existing and the new code) using `useMemo` for everything here? I'm pretty sure that you're using a more expensive operation (`JSON.stringify`) to save work on a simpler operation (`Array.filter`).

majewsky (approved):
Other than my performance concern above.

PR #1026, opened 2022-01-18
Email

sirajudheenam:


plugins/email_service/app/controllers/email_service/__configsets_controller.rb
```diff
@@ -0,0 +1,113 @@
+# module EmailService
```
ArtieReus, line 1: Do you really need to keep a copy of a controller? You can allways check the git history

plugins/identity/app/services/service_layer/identity_services/credential.rb
```diff
@@ -2,19 +2,52 @@
...
+          else
+            err = AWSCreds.new("" , "", "AWS EC2 credentials are not created. Without this, email service will not work. Open your web-console and execute `openstack ec2 credentials create` command")
+          end
+        else
+          err = "user_id / project can't be empty"
+        end
+        [err, creds]
+      end
+
+      def get_project(project_id, user_id)
```
ArtieReus: Please use the corresponding service layer method. Please do not recreate methods from other service layers, this could lead to errors. Use the ones defined here:
https://github.com/sapcc/elektra/blob/master/plugins/identity/app/services/service_layer/identity_services/project.rb

plugins/identity/app/services/service_layer/identity_services/credential.rb
```diff
@@ -2,19 +2,52 @@
...
+        res = elektron_identity.get("projects/#{project_id}/users/#{user_id}")
+      end
+
+      def get_user(user_id)
```
ArtieReus: Please use the corresponding service layer method. Please do not recreate methods from other service layers, this could lead to errors. Use the ones defined here:
https://github.com/sapcc/elektra/blob/master/plugins/identity/app/services/service_layer/identity_services/user.rb

ArtieReus (changes requested):

plugins/identity/app/services/service_layer/identity_services/credential.rb
sirajudheenam: Those were removed in the subsequent commit.

plugins/identity/app/services/service_layer/identity_services/credential.rb
sirajudheenam: Those were removed in the subsequent commit.

plugins/email_service/app/views/email_service/emails/index.html.haml
```diff
@@ -42,4 +42,4 @@
           recipients. 
           You can also send 
           %b
-            = link_to "templated eMails", "https://documentation.global.cloud.sap/services/email-howto-templates#send-template-email", :target => "_blank"
```
ArtieReus, line 45: Please still use the `:target => "_blank"`. We agreed that docu links should be open in a new tab.

plugins/email_service/app/models/email_service/plain_email.rb
```diff
@@ -0,0 +1,105 @@
+# module EmailService
```
ArtieReus, line 1: What is the point of keeping all this commented files?

ArtieReus (changes requested):
jus what I commented. The rest looks goold

plugins/email_service/app/views/email_service/emails/index.html.haml
sirajudheenam: # indicates the section of the html page. Github shows this (#) as comment. However It still opens in new tab.

plugins/email_service/app/models/email_service/plain_email.rb
sirajudheenam: Removed in the subsequent commit.

ArtieReus (approved):

PR #1031, opened 2022-01-26
fix plain_email to use ActiveModel and move ec2creds to concerns

sirajudheenam:
- Fixed PlainEmail to use ActiveModel for email validations using custom Validators.
- Moved ec2_creds to concerns
- WebController redirects a user without ec2 credentials to a page with instructions and embedded webconsole.
- Verified Domain items to clickable links.
- Corresponding Tests for PlainEmail Controller actions
- CoffeeScript - Added hidden click actions to fill fake data on forms for easier testing.


ArtieReus (approved):
lgtm

PR #1046, opened 2022-02-15
[compute instances][VMware vm tools] fix: prometheus query with instance_uuid

richardtief:
The prometheus API query uses the virtual instance `name`, which can cause errors because they are not unique - use the OpenStack `instance_uuid` instead. 

Screenshot 2022-02-15 at 17 22 05


hgw77 (approved):

hgw77:
LGTM 👍

PR #1047, opened 2022-02-21
add nebula support & ec2 credentials controllers

sirajudheenam:
* Added support to send email and templated email from domain
* Added nebula support
* Added ec2 credentials controller and dedicated tab for it.
* Added return path and reply to address


plugins/email_service/app/controllers/email_service/application_controller.rb
```diff
@@ -19,19 +19,58 @@ def restrict_access
         redirect_to index_path
       end
     end 
+
```
ArtieReus, line 22: wouldn't make more sence to do these checks with the policy? we used to use this way of checking roles because we wanted to redirec to to the old UI for testing purposes. All other roles check should go in the policy.

plugins/email_service/app/controllers/email_service/configsets_controller.rb
```diff
@@ -1,13 +1,14 @@
 module EmailService
   class ConfigsetsController < ::EmailService::ApplicationController
     before_action :restrict_access
-
+    before_action :check_user_creds_roles
```
ArtieReus, line 4: using the policy engine you wouldn't need this before checks

ArtieReus (changes requested):

plugins/email_service/app/controllers/email_service/application_controller.rb
sirajudheenam: Once we remove the role  `cloud_support_tools_viewer` in production, we need to check whether project has required roles and has credentials and `cronus` service is enabled or not. If one of them is missing it is supposed to render a page with hints and appropriate buttons to create credentials or enable cronus or add required roles. I think this check is required.

edda: The role checks are done by the policy engine if you have the policy configured correctly. You need to tell your controller `authorization_required` and then Elektra will show an error page stating the roles a user needs to be able to access the page without you having to do anything else. 

Also It is never a good idea to check for roles directly in a controller or view because you have to remember to change all the occurrences of the check if the role changes in any way.

Instead the correct check using the policy looks something like:
```
unless current_user.is_allowed?(email_service:email_list")
```

edda: One additional comment. In multi-layered boolean statements like the one below you should use parentheses to ensure the expression is evaluated the way you expect it to. So
```
unless (a && b) || (c && d)
```
And lastly, a double negative expression like the one above (starting with unless and then having a ! in the expression) is very hard to read and very easy to get wrong. I recommend simplifying it.

sirajudheenam: sure. I have changed those double negation and inserted brackets for conditions. Removed custom role checks and let default behaviour do its work. Thanks.

plugins/email_service/app/controllers/email_service/configsets_controller.rb
sirajudheenam: adjusted in further commits.

plugins/email_service/app/controllers/email_service/application_controller.rb
```diff
@@ -19,19 +19,58 @@ def restrict_access
...
+      if actions.include?(exception.resource[:action]) && controllers.include?(exception.resource[:controller])
         @title = 'Unauthorized'
         @status = 401
         @description = 'You are not authorized to view this page.'
         if exception.respond_to?(:involved_roles) && exception.involved_roles && exception.involved_roles.length.positive?
           @description += " Please check (role assignments) if you have one of the following roles: #{exception.involved_roles.flatten.join(', ')}."
         end
-        render '/email_service/shared/warning.html', status: @status
+        render '/email_service/shared/role_warning.html'
```
ArtieReus, line 69: Hi Sam! As I explained before elektra defaults to a standard behaviour where a page is shown with an error and the requested roles. You don't have to build your self an extra view for that. We have to try to keep the plugins consistent and looking the same. Have you tried in production mode and check what happens?

plugins/email_service/app/controllers/email_service/configsets_controller.rb
```diff
@@ -1,13 +1,14 @@
 module EmailService
   class ConfigsetsController < ::EmailService::ApplicationController
     before_action :restrict_access
-
+    before_action :check_ec2_creds_cronus_status
```
ArtieReus, line 4: Using the default behaviour you don't need to check here.

plugins/email_service/app/controllers/email_service/domain_verifications_controller.rb
```diff
@@ -1,6 +1,7 @@
 module EmailService
   class DomainVerificationsController < ::EmailService::ApplicationController
     before_action :restrict_access
+    before_action :check_ec2_creds_cronus_status
```
ArtieReus, line 4: same here

plugins/email_service/app/controllers/email_service/stats_controller.rb
```diff
@@ -1,16 +1,12 @@
 module EmailService
   class StatsController < ::EmailService::ApplicationController
     before_action :restrict_access
-
+    before_action :check_ec2_creds_cronus_status
```
ArtieReus, line 4: same here

ArtieReus (changes requested):

plugins/email_service/app/controllers/email_service/configsets_controller.rb
sirajudheenam: This is not checking roles anymore. It is checking whether the user has EC2 credentials and nebula enabled or not.

plugins/email_service/app/controllers/email_service/application_controller.rb
sirajudheenam: ok, thanks. removed the role checking and custom error handling.

plugins/email_service/app/controllers/email_service/stats_controller.rb
sirajudheenam: This is not checking roles anymore. It is checking whether the user has EC2 credentials and nebula enabled or not.

ArtieReus (approved):
lgtm! do not add anything more to this pr

PR #1062, opened 2022-04-08
[compute instances] new prometheus url on April 19, 9AM CET

richardtief:
We will offload the global `vrops_` metrics from `prometheus-infra-collector` starting April 19 at 9am CET. The data should then be queried by `prometheus-vmware`. No further changes are necessary - no certificates or the like. 

We will confirm in this PR when the move is complete.


richardtief:
@edda @andypf @ArtieReus @hgw77 

The migration is done!

edda (approved):

PR #1072, opened 2022-04-28
remove link from identity wizard to old broken resmgmt UI

majewsky:


majewsky:
@edda @andypf @ArtieReus  ping

andypf (approved):

PR #1080, opened 2022-05-19
[shared_filesystem_storage] bump microversion to max in xena release

Carthaca:
only share access rules need to stay at older 2.44 because this api got
deprecated and would require a rework

most interesting new feature comes with 2.63:
"Added the possibility to attach security services to share networks in use.
Also, an attached security service can be replaced for another one of the same 'type'.
In order to support those operations a 'status' field was added in the share networks
as well as, a new property called 'security_service_update_support' was included
in the share networks and share servers.
Also new action APIs have been added to the share-networks endpoint:
'update_security_service', 'update_security_service_check' and 'add_security_service_check'."

TODO: 
- [ ] implement add_security_service_check: to check if it is okay to add a security service to a share network that is in use
- [ ] implement update_security_service: to change parameters of a security service that is associated to a certain share network, along with the pre check update_security_service_check


Carthaca:
closing in favor of https://github.com/sapcc/elektra/pull/1183

PR #1128, opened 2022-09-30
audit module rename fields

notque:
Customer communicates the naming of fields not matching the actual events makes things confusing. I am not in love with all the name changes made here, but it is clearer. If you have any feedback, happy to get it.


IvoGoman (changes requested):
I would prefer the new names without their alternative. I also noticed the filter names and the table header in the UI do not match. Table headers are defined in L147-L151 in the changed file.

artherd42:
I would prefer the same names as defined in the CADF event: `observer`, `target` and `observer` also in the table header.

notque:
I've added what I took from the feedback. I feel like it's less clear, however it does match the CADF events now. The CADF events are less clear, so it just extends to that model. However, it's now very obvious what you're looking at. Is that an improvement?

IvoGoman (commented):
Yes. In my opinion it is now easier to get the connection between filters, table headers and the fields in the event. 

Not sure though if there are certain conventions w.r.t. column names. e.g. volumes uses `Volume Name` and the column contains the Name and the ID as a subscript. The `/` is used in the Servers page in the Name/AZ column. Here, it is used to distinguish two different attributes of a server in the same column. More a question for the Elektra Team.

edda:
I agree with @notque. The labels shown in the UI don't have to exactly match the keys in the API. In fact usually the very reason why graphical users interfaces are easier for people to understand is because the labels that are displayed have been made human readable instead of using the exact technical keys. In my opinion this change  doesn't make things necessarily easier to understand, especially not for people who never use the API themselves.

My suggestion would be to at least turn the `TypeURI` label into just `Type`. This is much more human-readable and I think the mental transfer from `Type` to `TypeURI` is something we can expect people who are used to the API keys to make.

notque:
I agree TypeURI should be Type, but I would go much stronger. This is a significantly worse user experience for customers, and no change would be better than the listed change in my mind.

@edda - Are there tool tips available reasonably in our structure? Could I add Tool Tips with the event's actual name?

hgw77:
what about this PR, is this still an issue? Or can we close it?

notque:
I’ll change it up to be the minimum change and we can get it in.

PR #1129, opened 2022-10-04
[lbaas2] Fix member weight range. Member weights can be 0.

BenjaminLudwigSAP:
Octavia pool members can have a weight between 0 and 256.


ArtieReus (approved):
lgtm

PR #1148, opened 2022-11-17
pull Juno dependency by tag version instead of by "latest"

majewsky:
This way, it should not randomly break when a new Juno version is tagged.


ArtieReus:
I am not really into this topic. @edda please review

majewsky:
FWIW I'm not attached to this PR at all. I remember Esther saying that she was also thinking about how to solve this. If you end up coming up with a better or just different solution, I'm all for it.

andypf (approved):

PR #1173, opened 2023-01-05
[lbaas] Make health monitor name attribute optional

BenjaminLudwigSAP:
When creating or editing a health monitor, the `Save` button can only be pressed if the `Name` field isn't empty.
However, [the name attribute for health monitors is optional](https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-health-monitor-detail#create-health-monitor).
This PR makes the `Name` field optional.

Note:
Health monitors without name can already be created via the CLI. Elektra then simply shows their ID instead of their name (under the label `Name/ID`). That means that Elektra already is (or at least has to be) able to handle this case. So the only thing that has to be changed is to make the `Name` field optional.

### DIFF
--- a/plugins/lbaas2/app/javascript/widgets/app/components/healthmonitor/EditHealthMonitor.jsx
+++ b/plugins/lbaas2/app/javascript/widgets/app/components/healthmonitor/EditHealthMonitor.jsx
@@ -199,7 +199,7 @@ const EditHealthMonitor = (props) => {
                 </p>
                 <Form.Errors errors={formErrors} />
 
-                <Form.ElementHorizontal label="Name" name="name" required>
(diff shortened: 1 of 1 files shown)


ArtieReus:
Hi @BenjaminLudwigSAP! Thank you for your proposal. The field name is required on purpose. All other objects in the lbaas plugin have the name also required. This is to keep consistency and better UI experience as default generally in elektra.

If there is no special requirements I would like to keep it like this.

ArtieReus:
The reason why the plug-in shows the id instead of the name is because of the CLI. This happens also with the other services. It is hard to display objects just with ids this is why in elektra we enforce names.

PR #1175, opened 2023-01-13
[lbaas] Health monitors: Fix mapping of max retries to API parameter

BenjaminLudwigSAP:
`max_retries` has been split up in [Octavia API](https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-health-monitor-detail#update-a-health-monitor) into `max_retries` and `max_retries_down`.
`max_retries_down` specifies the amount of retries for a HM before the pool member is considered down, while `max_retries` specifies the amount of retries for a HM before the pool member is considered up again.
We're only interested in the max retries before a member is down.

### DIFF
--- a/plugins/lbaas2/app/javascript/widgets/app/components/healthmonitor/EditHealthMonitor.jsx
+++ b/plugins/lbaas2/app/javascript/widgets/app/components/healthmonitor/EditHealthMonitor.jsx
@@ -233,7 +233,7 @@ const EditHealthMonitor = (props) => {
                     type="number"
                     min="1"
                     max="10"
-                    name="max_retries"
+                    name="max_retries_down"
                   />
                   <span className="help-block">
                     <i className="fa fa-info-circle"></i>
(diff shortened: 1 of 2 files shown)


ArtieReus:
Hi @BenjaminLudwigSAP thank you for the notice and the PR. I just have seen, that it should be also changed in other parts of the application as the HM model or the edit view, etc. I will check it next week.

BenjaminLudwigSAP:
(the force-push was to remove the change of Max Retries maximum value from 10 to 9. Turns out 10 is indeed the maximum value accepted by the Octavia API, while 9 is the maximum number accepted by the CLI... m(

ArtieReus:
@BenjaminLudwigSAP are you sure that you can submit a HM without `max_retries`. Exchanging the attribute to 'max_retries_down' would return following error:
```
Invalid input for field/attribute max_retries. Value: 'None'. Mandatory field missing.

The API description for the field `max_retries` doesn't say that it is optional.

BenjaminLudwigSAP:
Thanks for pointing this out. I suggest always supplying a fixed value of 1 for `max_retries`, since it is the amount of retries before changing a member to `ONLINE`, where 1 would adhere to the principle of least surprise imho.

BenjaminLudwigSAP:
Closed in favor of #1179

PR #1176, opened 2023-01-13
[lbaas] Health monitors: Remove Probe Timeout parameter

BenjaminLudwigSAP:
Probe Timeout (`timeout` on the API) is not honored by the Octavia provider driver. This is intentional. I understand this might be confusing, so an optional detailed technical explanation can be found below.

PR #1175 might seem similar to this, but it's not; #1175 fixes a mapping between Elektra parameter name and API parameter name, while this PR removes an unused parameter entirely.

Optional detailed technical explanation:
Health monitors on the F5 backend only accept two parameters: `interval` and `timeout`. While the `interval` backend parameter has the same semantics as the `delay` Octavia API parameter (that is, the amount of seconds between probes), the `timeout` backend parameter has slightly different semantics than the Octavia API `timeout` parameter: While the `timeout` Octavia API parameter denotes the timeout *of a single probe*, the `timeout` backend parameter denotes the timeout `over all probes`. Therefore the `timeout` Octavia API parameter cannot be used here. Instead we calculate the `timeout` backend parameter (that is, the overall timeout) as the value of the `max_retries_down` Octavia API parameter (that is, the number of failing probes before the member is set to OFFLINE) multiplied by the `delay` Octavia API parameter (that is, the amount of seconds between probes) plus one spare second to give the last probe a chance to not run into a timeout. That calculation happens here: https://github.com/sapcc/octavia-f5-provider-driver/blob/stable/yoga-m3/octavia_f5/restclient/as3objects/monitor.py#L122

### DIFF
--- a/plugins/lbaas2/app/javascript/widgets/app/components/healthmonitor/HealthmonitorDetails.jsx
+++ b/plugins/lbaas2/app/javascript/widgets/app/components/healthmonitor/HealthmonitorDetails.jsx
@@ -114,12 +114,12 @@ const HealthmonitorDetails = ({ loadbalancerID, poolID, healthmonitor }) => {
       <div className="list-entry">
         <div className="row">
           <div className="col-md-12">
-            <b>Retries/Probe Timeout/Interval:</b>
+            <b>Retries/Interval:</b>
           </div>
         </div>
         <div className="row">
           <div className="col-md-12">
-            {healthmonitor.max_retries} / {healthmonitor.timeout} /{" "}
+            {healthmonitor.max_retries} /{" "}
             {healthmonitor.delay}
           </div>
         </div>
--- a/plugins/lbaas2/app/javascript/widgets/app/components/healthmonitor/EditHealthMonitor.jsx
+++ b/plugins/lbaas2/app/javascript/widgets/app/components/healthmonitor/EditHealthMonitor.jsx
@@ -242,25 +242,6 @@ const EditHealthMonitor = (props) => {
                   </span>
                 </Form.ElementHorizontal>
 
-                <Form.ElementHorizontal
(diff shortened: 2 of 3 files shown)


plugins/lbaas2/app/javascript/widgets/app/components/healthmonitor/HealthmonitorDetails.jsx
```diff
@@ -114,12 +114,12 @@ const HealthmonitorDetails = ({ loadbalancerID, poolID, healthmonitor }) => {
...
         <div className="row">
           <div className="col-md-12">
-            <b>Retries/Probe Timeout/Interval:</b>
+            <b>Retries/Interval:</b>
           </div>
         </div>
         <div className="row">
           <div className="col-md-12">
-            {healthmonitor.max_retries} / {healthmonitor.timeout} /{" "}
+            {healthmonitor.max_retries} /{" "}
```
BenjaminLudwigSAP, line 122: I'm not sure what the semantics of `/{" "}` are exactly, so please check that my removal is correct, thanks!

ArtieReus:
Hi @BenjaminLudwigSAP Thx for the info. I will take care as soon as possible. It can't be merged directly since the model of the object and tests habe to be changed also.

ArtieReus:
hi @BenjaminLudwigSAP. Same as in PR #1175. Did you test to create a HM without the attr `timeout` because I am getting following error:

```
Invalid input for field/attribute timeout. Value: 'None'. Mandatory field missing.
```

BenjaminLudwigSAP:
Same solution then: Supply a fixed value, e.&nbsp;g. 0. As I mentioned in the original post, the `timeout` value isn't honored by Octavia provider driver anyway.

PR #1182, opened 2023-01-26
[manila] change listing share access rules from POST to GET

Carthaca:
new method was added with microversion 2.45, see
https://docs.openstack.org/api-ref/shared-file-system/#list-share-access-rules

benefit: since audit middleware ignores GET requests, we get less audit spam

- [x] tested with local dev env in qa


edda (approved):
Since you have tested it I trust you :D

PR #1195, opened 2023-02-21
[compute] Update nova policy for Xena upgrade

grandchild:
As discussed in Slack, wait with merging until all regions have been upgraded.

> The worst that will happen is that the policy mismatch disallows an action in Nova that was still allowed from Elektra side.


edda:
I am pretty sure that this will not work. It looks like you copied the nova policy in here but that's not how it works. In Elektra the policy must be mapped to controller actions. Often times the controller actions correspond 1:1 to API calls but not always. Look for example at the instance actions. In Elektra they follow the pattern `compute:instance_action` in the Nova policy it seems to be `compute:action`. 

So to recap: if you change the policy "key" (first part of the policy statement before the colon) the policy won't be evaluated properly anymore. In addition I noticed some networking policies in there. If there are changes there those need to go to the networking plugin in Elektra not to the compute plugin.

hgw77:
can this PR be closed?

PR #1197, opened 2023-02-24
Update react-query to v4

ArtieReus:


package.json
```diff
@@ -62,6 +62,7 @@
     "redux": "^4.0.0",
     "redux-devtools-extension": "^2.13.5",
     "redux-thunk": "^2.2.0",
+    "sass": "^1.58.3",
```
ArtieReus, line 65: ist sass eine dependency vom react-query v4?

ArtieReus (commented):
@hodanoori kann you please also change all react-query hooks to use an object. React-query prefers object with option, as it is more consistent with other React Query hooks, like useQueries. It will also be the only available option in a future major version.

PR #1226, opened 2023-06-19
keppel: change sort order of the contains column to prefer images over tags

SuperSandro2000:
This is more logical since images are the first number in the row and also use at least a tiny amount of storage while tags are almost free.

@majewsky what do you think?


majewsky (approved):

hgw77 (approved):
👍

PR #1254, opened 2023-08-16
allow edit tag with compute-admin

ZhichaoTJ:
our customer recently noticed that without keystone admin role, he is not able to see "Edit Tag" button, which does not feels right. A common sense is anyone with compute_admin role should be able to edit tags. After examine the code, I propose the change in this PR, please kindly review.

image
image


ZhichaoTJ:
meant to create a draft PR. Let me close this one.

PR #1292, opened 2023-11-17
[keymanagerng] Download payload file as a raw file not a json file 

hodanoori:
and show the payload content directly in the secret details panel if the content type is text/plain

### DIFF
--- a/plugins/keymanagerng/app/javascript/widgets/app/components/secrets/secretDetails.jsx
+++ b/plugins/keymanagerng/app/javascript/widgets/app/components/secrets/secretDetails.jsx
@@ -36,6 +36,7 @@ const SecretDetails = () => {
   const history = useHistory()
   const params = useParams()
   const [secretId, setSecretId] = useState(null)
+  const [payloadString, setPayloadString] = useState("")
   const [creatorName, setCreatorName] = useState(null)
   const [secretMetadata, setSecretMetadata] = useState(null)
   const [payloadRequested, setPayloadRequested] = useState(false)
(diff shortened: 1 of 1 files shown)


plugins/keymanagerng/app/javascript/widgets/app/components/secrets/secretDetails.jsx
```diff
@@ -88,18 +89,44 @@ const SecretDetails = () => {
       secret?.data?.content_types?.default,
     ],
     queryFn: getSecretPayload,
-    enabled: !!payloadRequested,
+    enabled:
+      payloadRequested || // Enable when download requested
+      (![
```
edda, line 94: I find this boolean expression a bit hard to read, it's easy to miss the `&& show` at the end. My suggestion:
```javascript
payloadRequested  || (show && ![....].includes(...))
```

plugins/keymanagerng/app/javascript/widgets/app/components/secrets/secretDetails.jsx
```diff
@@ -88,18 +89,44 @@ const SecretDetails = () => {
...
+        show), // Enable for specific content types when panel is shown
     onSuccess: (data) => {
-      const fileData = JSON.stringify(data)
-      const blob = new Blob([fileData], { type: "text/plain" })
-      const url = URL.createObjectURL(blob)
-      const link = document.createElement("a")
-      link.download = secret?.data?.name + ".json"
-      link.href = url
-      link.click()
+      if (
```
edda, line 101: Similar to above, it's easy to miss the not (`!`). Maybe change the if so that it doesn't check for what it is not, but what it is. So something like:
```javascript
if([...].includes(...)) {
  // Handle specific payload types by....
...
} else {
  // Handle payload types other than ...
...
}
```

plugins/keymanagerng/app/javascript/widgets/app/components/secrets/secretDetails.jsx
edda: One additional suggestion. You use
```javascript
[
                      "application/octet-stream",
                      "application/pkcs8",
                      "application/pkix-cert",
                    ].includes(secret?.data?.content_types?.default) 
```
three times here. I think it would be much more readable if you wrote a little helper function `const isDownloadableDataType`

Then you can do e.g. 
```javascript
payloadRequested  || (show && !isDownloadableDataType(secret?.data?.content_types?.default))
```

edda (changes requested):

plugins/keymanagerng/app/javascript/widgets/app/components/secrets/secretDetails.jsx
hodanoori: Fixed

edda (approved):

PR #1295, opened 2023-11-23
[keymanagerng] Show payload download button for all content types

hodanoori:
and 

- [SecretDetails] Show the first 500 characters of payload in a codeblock for text/plain content types
- [New Secret] Show expiration time correctly which is always 23:59:59
- [New Secret] Show secret types info as helptext


ArtieReus (approved):
lgtm

PR #1307, opened 2024-01-18
prepare UI strings for upcoming new compute quota structure

majewsky:
There will be more quotas, so I will be splitting them into more categories to keep things organized. These are the category names.


hgw77 (commented):
👍

PR #1367, opened 2024-06-12
[octavia] - add xfh insert header option to listener

m-kratochvil:
In this PR I'm adding additional option to the `Insert Headers` drop down under the load balancer listener configuration dialog.
The option name is `X-Forwarded-Host` and the option is applicable to listener with protocol type `HTTP` or `TERMINATED_HTTPS`. The implementation should be the same as with the existing `Insert Headers` options, for example the "X-Forwarded-For".

I am also adjusting wording of the `description` of the other X-Forwarded-xxxx options.


m-kratochvil:
Converting to draft as I found an issue with API validation for this new option, which I have to solve first

m-kratochvil:
Will not be implemented this way. Closing PR, new one will be opened later.

PR #1371, opened 2024-06-20
plugins/identity: remove obsolete integrations with Resource Mgmt plugin

VoigtS:
This PR aims to remove deprecated integrations of the old Resource Management Plugin that do not exist anymore in the new quota/commitment model.

The following removals take place:
- wizard step for setting up quota (is now assigned automatically)
- information about status of quota bursting (has been removed)

We executed the corresponding ruby spec file: `./plugins/identity/spec/controllers/projects_controller_spec.rb`. We also took a look at the UI behaivior after the changes. We don't expect any issues with the removal of the above mentioned parts.


plugins/identity/spec/controllers/projects_controller_spec.rb
```diff
@@ -182,45 +159,6 @@
       end
     end
 
-    context "masterdata_cockpit and networking services are not available" do
```
hgw77, line 185: it is ok to remove the whole spec? because it removes also tests for masterdata, networking, sharding

VoigtS: You are right, that was too much. Now only the resource_management related test is removed, as originally intended.

PR #1383, opened 2024-07-10
Remove old resources plugin components

VoigtS:
With this PR I want to remove the old resources plugin from elektra, so that we only have the new `Resource Management` plugin left in the system.

What needs to be cleared up first(discussed):
* I left the `resources/services/service_layer/resources_service.rb` in tact, because it contains bigVM methods. The methods defined there do still receive changes. However, I can't seem to find any method calls of them in the remaining code base. So I need to know what can be removed there and what can not.
* ~We still have a limes subplugin in the `tools` plugin, as well as Castellum. But they are not displayed in the UI anymore. Can those be removed too?~ I talked with Stefan Majewsky about it. They are referenced in prometheus alerts. I added comments for clarity.

What I removed:
* The core JS application
* From now on obsolete controllers of the plugin
* The links and references that lead to the old UI.

What stayed:
* The necessary parts (like the gemspec and policy configs) to keep the the new UI running
* The bigVM controller that defines the HANA flavor endpoints.

Additional Fix:
* An identation error in the identity plugin caused the `network` wizard to not run at all. This caused a infinite loading bug on project configurations.


app/assets/builds/.keep
SuperSandro2000, line 1: You probably deleted this file by accident when we tried to fix the caching issue you had yesterday.

edda (approved):
lgtm

majewsky:
Re `plugins/resources/app/services/service_layer/resources_service.rb`: This is all dead code, its usage was removed in 97d2928af6a31dc86726663b0c61908bc2815a21

plugins/resources/config/routes.rb
```diff
@@ -5,33 +5,8 @@
   # NOTE: "current" used to be ":cluster_id". Since multi-cluster support has
   # been removed from the UI, the value "current" is hardcoded to preserve
   # backwards compatibility.
-
-  get "/project/current/:override_domain_id/:override_project_id" =>
-        "application#project",
-      :as => "foreign_project"
-  get "/project" => "application#project", :as => "project"
-  get "/project/bigvm_resources" => "application#bigvm_resources",
-  :as => "bigvm_resources"
```
majewsky, line 14: This is used by the new UI and needs to stay.

majewsky (approved):
Looks good otherwise. I vote that we move ahead with this now, since there is still more stuff to clean up in `plugins/resource_management`.

plugins/resources/config/routes.rb
VoigtS: That was an oversight. Thank you.

majewsky (approved):

PR #1384, opened 2024-07-12
maillog-ui (juno app) view added as tab to email-service

ornokreans:
A view for maillog UI added as a tab to Email-service plugin.
maillog-ui is a juno app - https://github.wdf.sap.corp/cronus/maillog-ui


andypf (approved):
lgtm

PR #1389, opened 2024-07-23
Fix e2e tests for resources

VoigtS:
This should be the last adjustment that needs to be done from the origin PR: #1383
We have failing e2e tests for the resources plugin, because it lead to the old URI. We now route to the `v2/` path with the new page title that will be checked for.

For clarification: I removed the check for a specific resource here, because the UI now only shows committable resources per default and hides the others behind a "show" setting. Those can change any time, so this check might break.

Edit: I removed the project level resources tests completely -> They checked for specific resource names, we want to decouple the new UI from that.
We also fixed swift. The old project path that was still there caused its UI to crash. This fixes the remaining e2e tests that currently fail.


andypf (approved):
LGTM

PR #1393, opened 2024-08-06
fix(instances): disable project fip cache cache

### LINKED ISSUE #1385: Deprecate nova `os-floating-ips` calls to show the VM FIP on the servers page
The https://docs.openstack.org/api-ref/compute/?expanded=list-hypervisors-detail%2Ccreate-server-detail#list-floating-ip-addresses call has been deprecated and it was also deprecated in gophercloud v2, which is used in terraform.

Now the FIP association must be done directly with the server's port ID. And with this change Elektra doesn't show the FIPs associated with the VM. Once a new terraform provider version will be released, users will start to complain on the inconsistency between terraform and UI.

See:

* https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/compute_floatingip_v2
* https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs/resources/compute_floatingip_associate_v2
* `openstack/compute/v2/extensions/floatingips` - This was a proxy for the Networking service, Neutron. Use `openstack/networking/v2/extensions/layer3/floatingips` instead - https://github.com/gophercloud/gophercloud/blob/master/docs/MIGRATING.md

There are two ways to identify the FIP <-> Server association:

* get `port_id`'s listing the ports with `?device_id=%server_id%`, then list `floating_ip`'s with the corresponding `?port_id=%port_id%&port_id=%port_id%`
* list `floating_ip`'s using the server's `fixed_ip`'s: `?fixed_ip_address=1.2.3.4&fixed_ip_address=4.3.2.1`. But the latter solution is less precise since there could be different networks with the same `fixed_ip_address`'es.

hgw77:
this will fix #1385


plugins/compute/app/helpers/compute/instances_helper.rb
```diff
@@ -349,4 +357,4 @@ def render_fixed_floating_ips(ips)
     # End op Floating IPs
     #########################################################################
   end
-end
+end
```
andypf, line 360: please add a new line here

plugins/compute/app/models/compute/server.rb
```diff
@@ -500,4 +587,4 @@ def validate_network
       errors.add(:network_ids, "Please select a network")
     end
   end
-end
+end
```
andypf, line 590: new line

plugins/compute/app/views/compute/instances/index.html.haml
```diff
@@ -54,4 +54,4 @@
   .ajax-paginate{ data: { ajax_paginate: false, list_selector: 'table.instances tbody', search_input_selector: '#search_instances_input'} }
 
 
-= render_paginatable(@instances, {searchfor: @searchfor, search: @search})
\ No newline at end of file
+= render_paginatable(@instances, {searchfor: @searchfor, search: @search}, {disable_show_all: true})
```
andypf, line 57: same here

plugins/compute/app/helpers/compute/instances_helper.rb
```diff
@@ -298,23 +298,31 @@ def network_ips_map(ips)
...
+        #instance_floating_ips = instance.floating_ips.collect { |f| f["addr"] }
+
+        #puts (available_floating_ips & instance_floating_ips).sort
+        #puts instance_floating_ips.sort
+
+        # check if there are floating ips that are not in the cache
+        #if (available_floating_ips & instance_floating_ips).sort != instance_floating_ips.sort
+        #  puts '::::::::::::::::::: REFRESH CACHE ::::::::::::::::::'
+          @project_floating_ips = services.networking.project_floating_ips(@scoped_project_id)
+        #end
```
andypf, line 325: Do we need this commented-out code?

hgw77: it's for quick reference if the solution without cache is not working sufficiently, It can be removed but I would wait some time to prove it

hgw77:
https://marketplace.visualstudio.com/items?itemName=github.vscode-codeql

PR #1400, opened 2024-08-23
feat(object store): handle 10k limit for container object listing

hgw77:
this PR will handle the 10000 files limit.

Background: the swift api will only report max 10k files in the list call.

Solution: check if we have 10k of files in the list. If yes we try to load the next with the `marker` option until the file count is below 10k


andypf (approved):
LGTM

Issue #1395, opened 2024-08-06
fix(elektra/security_groups) implement search and pagination for security_groups list


ArtieReus:
#### Description

The Elektra plugin for security groups currently lacks pagination when listing items. This has become a significant issue as the some projects have grown quite a lot. 

As an example, we have more than 750 security groups in follwoing project, but we can only list and search within the first 500 items.

[Slack issue discussion](https://convergedcloud.slack.com/archives/C2WPEPV8B/p1722923299607159)
[Slack API limit discussion](https://convergedcloud.slack.com/archives/CAVMRU0BB/p1722944927658519)

#### Problem

- No pagination is implemented.
- Some projects have already over 750 security groups.
- Only the first 500 items are accessible for listing and searching.

#### Proposal 1

- Render a maximum of 20 items at a time.
- Implement infinite scrolling to display more items as the user scrolls.
- Fetch more items dynamically during scrolling or searching and keep them in memory.

#### Proposal 2

- Implement conventional pagination with 20 items per page.
- Use buttons to navigate between pages.
- Keep fetched items in memory.
- Implement smart searching that works with pagination.

#### Conclusion

Adding pagination will improve usability and ensure that all security groups are accessible for listing and searching. Please consider the above proposals for implementation.

PR #1408, opened 2024-09-03
fix(security groups): handle 500 items limit

hgw77:
# Summary

will load all security groups beyond the 500 items limit

this is a fix that works without larger rewrite of the code, just use the marker option to load all items on server side, than the user can use the search on client side to find the item. 

But it comes with a grain of salt 

1. it can be a problem if more than 1 or 2k Security Groups are found
2. the Api is not that fast to deliver the items if the count is beyond 500 

# Changes Made

- adjust the loading funktion in the controller to load all security groups with marker if more than 500 are found
- adjust view to show the count of all security groups
- disable the new button if loading is in progress
- add api docu to service layer

# Related Issues

- Issue 1: #1395 

# Checklist

### DIFF
--- a/plugins/networking/app/controllers/networking/security_groups_controller.rb
+++ b/plugins/networking/app/controllers/networking/security_groups_controller.rb
@@ -7,10 +7,20 @@ module Networking
     authorization_required
 
     def index
-      security_groups = services.networking.security_groups
-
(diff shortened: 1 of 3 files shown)


andypf (approved):
LGTM

PR #1413, opened 2024-09-12
remove all remaining usages of the old @quota_data mechanism

majewsky:
These were partially already broken because Juno plugins do not see @quota_data, or they are utterly superfluous because there is no action the user can take to mitigate low quota anyway.


hgw77 (commented):
👍

PR #1414, opened 2024-09-12
remove plugins/resource_management

majewsky:
As far as I can tell, there are no references left to this plugin anywhere else. The only usage of the word `resource_management` that remains is as a section identifier in the service navigation:

```
$ rg -A2 resource_management
config/navigations/services_navigation.rb
416:    primary.item :resource_management,
417-                "Capacity, Masterdata & Metrics",
418-                nil,
```

**Important question:** Do we know if it is safe to delete DB migrations like this? Will it confuse the migration handler if there are existing entries in `schema_migrations` that do not correspond to migrations defined in the code?


hgw77 (approved):

hgw77:
About the `schema_migrations` We need to investigate the existing entries in the schema_migrations table. To address this issue, we should create an additional migration to remove outdated or problematic entries from the database. This will help us clean up the migration status and prevent future conflicts. I will take look.

hgw77 (changes requested):
About the schema_migrations We need to investigate the existing entries in the schema_migrations table. To address this issue, we should create an additional migration to remove outdated or problematic entries from the database. This will help us clean up the migration status and prevent future conflicts. I will take look.

hgw77 (approved):
Upon investigation, it appears that the tables `resource_management_resources` and `resource_management_capacities` have already been deleted from the database. A migration in the resource_management folder handles this deletion process, and when examining the database migrations in the development environment of Elektra with the latest source, I observed that this migration runs successfully on newly created databases. As a result, no `resource management` database is found after the migration process. 



To provide context, the migration to drop these tables was executed in the production environment back in spring 2017 (we can guess that from the timestamp of the drop table migration file). Since then, these tables have been absent from the database. Given this information, we can safely remove the database migration files related to resource management. These files are now obsolete and no longer serve a purpose in our current database [schema](https://github.com/sapcc/elektra/blob/master/db/schema.rb).

majewsky:
Yes, I already saw that the tables were cleaned up. My concern was for the entries in `SELECT * FROM schema_migrations`. If you feel like we are fine on that front, please feel free to merge when you're ready.

Issue #1446, opened 2024-10-30
chore(cypress): repair tests


hgw77:
some tests are no more working because of changes to the landing page or upgrades

PR #1447, opened 2024-10-30
chore(cypress): fix tests

hgw77:
# Summary

* this will fix some failing tests

# Related Issues

#1446 


e2e/cypress/integration/admin/domain.js
```diff
@@ -23,8 +19,7 @@ describe("domain landing page", () => {
     // eslint-disable-next-line cypress/no-unnecessary-waiting
     cy.wait(500)
     // check not in one string because it can be different order
-    cy.contains("SAP Converged Cloud")
-    cy.contains("button", "Log in")
+    cy.contains("button", "Enter CC3TEST")
```
ArtieReus, line 22: Would be better hier to check for an fix button.id instead with the button text? If we change the button text the test would then fail...

e2e/cypress/integration/member/dns.js
```diff
@@ -22,14 +18,12 @@ describe("dns", () => {
     cy.contains("Request New Domain")
     cy.get("#zone_request_domain_pool").select("Internal SAP Hosted Zone")
     // click Subdomain
-    cy.get("input#zone_request_domain_type_subdomain")
-      .should("be.visible")
-      .click()
-    cy.get("input#zone_request_name").should("be.visible")
+    //cy.get("input#zone_request_domain_type_subdomain")
```
ArtieReus, line 21: We might habe to remove commented code

ArtieReus (commented):
This pull request looks good, but wouldn't be a good chance to improve it using element ID instead of generally looking for button text to check or do an action?

e2e/cypress/integration/admin/domain.js
hgw77: this is on purpose to check that the automatic generation of the button title is working, the test is running on `CC3TEST` domain so we should see the button with this text

ArtieReus: I was meaning generally in the tests. But in this specific one I would do following:

```js
cy.get("#yourButtonID").should("have.text", "Enter CC3TEST");
```

andypf (approved):
LGTM

PR #1448, opened 2024-10-30
Add minimum is critical feature to the UI

VoigtS:
# Summary

This is a manila UI change. Our customers desire the possibility to immediately scale their volume size if the option "Ensure this much free space" is set within the UI in order to solve sizing bottlenecks within the autoscaler. This PR implements a checkbox below this option which is only enabled if a input in the minimum size is present and attaches a attribute to the config that is then sent to the API.

# Screenshots (if applicable)
image

# Checklist

### DIFF
--- a/plugins/shared_filesystem_storage/app/javascript/widgets/app/components/castellum/configuration/edit.jsx
+++ b/plugins/shared_filesystem_storage/app/javascript/widgets/app/components/castellum/configuration/edit.jsx
@@ -9,13 +9,8 @@ const FormBody = ({ close, errors }) => {
   const values = useContext(Form.Context).formValues
 
   let validationError = ""
-  if (
-    values.low_enabled == "false" &&
-    values.high_enabled == "false" &&
-    values.critical_enabled == "false"
-  ) {
-    validationError =
-      "To use autoscaling, at least one threshold must be enabled."
+  if (values.low_enabled == "false" && values.high_enabled == "false" && values.critical_enabled == "false") {
+    validationError = "To use autoscaling, at least one threshold must be enabled."
   }
 
   return (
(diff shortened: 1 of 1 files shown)


plugins/shared_filesystem_storage/app/javascript/widgets/app/components/castellum/configuration/edit.jsx
```diff
@@ -190,6 +146,17 @@ const FormBody = ({ close, errors }) => {
...
           <Form.Input elementType="input" type="number" min="0" />
+          <div className="checkbox">
+            <label>
+              <Form.Input
+                elementType="input"
+                type="checkbox"
+                name="minimum_free_is_critical"
+                disabled={values.free_minimum == ""}
+              />
+              <span>Treat crossings of this constraint as a critical usage level</span>
```
edda, line 157: To me it's not immediately clear what this text means but I'm also not deeply into the topic. Will our users understand what this means?

majewsky: This is the best phrasing that I can come up with that ought to be clear while fitting in a single line. The word "critical" refers to a few lines above in the same form, where users configure low/high/critical thresholds.

andypf: I suggest: `If usage crosses this threshold, mark it as a critical level.`

majewsky: I don't see how this is better.

VoigtS: I now rephrased it to Stefans suggestion: `When crossing this threshold, resize without delay` which should be more clear now.

andypf (approved):
LGTM 👍

hgw77 (approved):

PR #1451, opened 2024-11-01
Add clear resize error action to admin UI

VoigtS:
# Summary

Errors in Castellum that are from the type "resize" do not get immediately resolved when the error disappears. It only disappears after another resize attempt on the resource occurs. For this case a new API endpoint was introduced that clears an error from the database early. The follwing changes are present:

- Add new Action column to the resize error table
- Content is a "Clear" button that aims to remove the error from the error list.
- Implement the API endpoint and the caller function
- Refetch existing errors after calling the "Clear" functionality.

# Screenshots (if applicable)

image

# Checklist

### DIFF
--- a/plugins/tools/app/javascript/widgets/castellum/components/error_list.jsx
+++ b/plugins/tools/app/javascript/widgets/castellum/components/error_list.jsx
@@ -131,7 +131,7 @@ export default class Loader extends React.Component {
         pageSize={6}
       >
         {data.map((error, idx) => (
-          <ErrorRow key={`error${idx}`} error={error} action={{label: "Clear", fn: this.onAction}}/>
+          <ErrorRow key={`error${idx}`} error={error} action={{label: "Mark as resolved", fn: this.onAction}}/>
         ))}
       </DataTable>
     )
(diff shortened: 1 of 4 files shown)


plugins/tools/app/javascript/widgets/castellum/components/error_list.jsx
```diff
@@ -118,7 +131,7 @@ export default class Loader extends React.Component {
         pageSize={6}
       >
         {data.map((error, idx) => (
-          <ErrorRow key={`error${idx}`} error={error} />
+          <ErrorRow key={`error${idx}`} error={error} action={{label: "Clear", fn: this.onAction}}/>
```
majewsky, line 134:
```suggestion
          <ErrorRow key={`error${idx}`} error={error} action={{label: "Mark as resolved", fn: this.onAction}}/>
```
"Clear" sounds like it's performing an action on the asset, instead of merely updating the presented state.

majewsky (changes requested):

plugins/tools/app/javascript/widgets/castellum/components/error_list.jsx
VoigtS: I checked how the longer name behaves in conjunction with the column size and it seems to be okay for the other content to still fit in the row.

majewsky (approved):

hgw77 (approved):

PR #1482, opened 2024-12-10
fix(keppel): reduce account deletion to necessary steps

VoigtS:
# Summary

Keppel was updated regarding its account deletion handling. Most of the heavy lifting is now handled via separate goroutines. This PR updates the UI to match the new account deletion handling for Keppel.

* remove all `in_maintanance` references from the UI. This is not managed anymore. Instead every info that the UI needs will be provided via a `state` attribute.
* The `status` column now displays an account as `<spinner> Deleting...`, because the deletion itself is now handled asynchronously by the backend. No actions can be performed for this account at this state. This required to modify the corresponding Subroute contents to return early.
* Multi Step deletion was removed. The API either provides an OK or an Error to the request.
* Multiple little bug fixes: 
	* The API errors werent properly displayed previously, because React was not imported. 
	* No parameter was provided to the `handleDoneDeleting` function. This caused success messages to never pop up.
	* The account delete cancelling was never catched, which always threw an error to the console.

# Screenshots (if applicable)

image

# Checklist


plugins/keppel/app/javascript/widgets/app/components/utils.jsx
```diff
@@ -63,3 +63,10 @@ export const makeSelectBox = ({ options, value, isEditable, onChange }) => {
 
 const trimEllipsis = (str) =>
   str.substr(-3) === "..." ? str.substr(0, str.length - 3) : str
+
+export const apiStateIsValid = (state) => {
+  if (state == 'deleting') {
+    return false
+  }
+  return true
+}
```
majewsky, line 72: This is a strange function name. The state `deleting` is valid.
```suggestion
export const apiStateIsDeleting = (state) => state === 'deleting'
```
Also, it is probably clearer at the callsites what is happening to write `state === SOME_CONSTANT` with a constant for the value `'deleting'` declared here or somewhere else.

VoigtS: I was not happy with the naming either and wanted to change it this morning too. The change is in: [2bc61f1](https://github.com/sapcc/elektra/pull/1482/commits/2bc61f198b59ec5c5af4084e266a4efd11073fbf)
I wanted to keep the function more broad at first to include additonal cases in the future, however having one clear state for it is preferable for me too here.

hgw77 (approved):

majewsky (approved):

PR #1484, opened 2024-12-12
chore(deps): bump limes ui

VoigtS:
# Summary

LimesUI version bump to 1.3.4


ArtieReus (commented):
Hi @VoigtS ! why not upgrading  also `juno-ui-components`?

VoigtS:
Updating the UI components will happen with the next release.
This one fixes the behavior of the feature for short term commitments.
For the UI components I now have time to check if anything breaks. But I don't want to include it into the same fix.

hgw77 (approved):

PR #1494, opened 2025-01-20
feat(masterdata-cockpit): add propagation information

hgw77:
# Summary

will add propagation type information to masterdata


plugins/masterdata_cockpit/lib/masterdata_cockpit.rb
```diff
@@ -63,5 +63,12 @@ def self.business_criticalitys
         "prod_tc" => "Production Time Critical",
       }
     end
+
+    def self.propagation_role_types
```
andypf, line 67: Why do you use a map where the key is an index?

andypf (approved):

PR #1514, opened 2025-02-05
fix(core): defer avatar img loading to unblock auth_projects fetch

edda:
## Summary

This PR refactors the `loadAvatar` function to ensure it is **truly non-blocking** and does not interfere with other asynchronous operations, such as fetch requests.

Even though the original implementation wrapped image loading in a Promise to try to unblock other requests, this still unintentionally deferred other API calls due to browser network prioritization (the browser tried to load images first). In our case this led to the `auth_projects` call being deferred until after the image load was completed or failed.

This updated implementation leverages `requestIdleCallback()` to defer image loading until the browser is idle, ensuring that fetch requests are not delayed.

### 1. **Removed Unnecessary Promise Wrapping**
   - The original implementation used a Promise to handle image loading.
   - Since image loading is already **event-driven** (`onload` / `onerror`), wrapping it in a Promise was unnecessary.
   - The new implementation **removes the Promise**, reducing overhead.

### 2. **Eliminated `insertAdjacentHTML()` in Favor of Direct DOM Manipulation**
   - Previously, a new `` element was created using `insertAdjacentHTML()`, which caused **duplicate element creation**.
   - The new approach **inserts the preloaded image element directly** using `appendChild()`

### 3. **Deferred Image Loading Using `requestIdleCallback()`**
   - In the original version, calling `img.src = avatarUrl` immediately triggered an image request, which could **block or delay fetch() requests** (particularly noticeable for `auth_projects`).
   - The updated version **uses `requestIdleCallback()`** to start loading images **only when the browser is idle**.
   - This ensures that **critical fetch requests complete first**

### 4. **Added a Timeout to Ensure Images Load Within a Reasonable Time**
   - If the browser is never idle (e.g., due to user interactions), `requestIdleCallback()` might delay execution indefinitely.
   - A **`timeout: 5000` (5 seconds)** ensures that the image still loads **even if the browser is under constant load**.

### 5. **Preserved `DOMContentLoaded` Handling**
   - If `loadAvatar` is called before the DOM is ready, it waits for `DOMContentLoaded` before executing.
   - If the DOM is already loaded, it runs immediately.

### 6. **CSS improvements**
  - Unrelated to the above fix, improved avatar css by reducing repetition of width/height setting.

# Checklist

### DIFF
--- a/app/assets/stylesheets/_monsoon_theme.scss
+++ b/app/assets/stylesheets/_monsoon_theme.scss
@@ -1058,12 +1058,9 @@ section {
 .avatar {
   background: inline_svg("avatar_default") no-repeat;
   background-size: contain;
-  height: 24px;
-  width: 24px;
-
   margin-right: 3px;
 
-  img {
+  &, img {
     border-radius: 3px;
     height: 24px;
     width: 24px;
@@ -1071,9 +1068,7 @@ section {
 }
 
 .avatar_profile {
-  height: 140px;
-  width: 140px;
-  img {
+  &, img {
     height: 140px;
     width: 140px;
   }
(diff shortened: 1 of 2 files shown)


hodanoori (approved):

andypf (commented):
👍

Issue #1517, opened 2025-02-10
[EPIC](secrets): Secrets Rotation


ArtieReus:
**Description**
This epic will track tasks, ideas, processes, and solutions to transition Elektra’s secret management to a new approach that enables seamless rotation while maintaining system stability and compliance. The goal is to implement a process that allows secrets to be updated without downtime, ensuring continuous operation. Details of the implementation and internal documentation will be created in our [Secrets Documentation](https://ui-team.global.cloud.sap/docs/house_keeping/secrets/) section.

**Tasks**

- [x] #1518
- [x] #1496
- [x] #1525
- [x] #1530
- [ ] \[TASK](secrets): use application credentials from the seeder
- [x] #1547
- [x] extend documentation: https://github.wdf.sap.corp/cc/documentation-operation/pull/2503/files for rails token
- [x] Add alerting.
- [x] Rotate smtp credentials manually.
- [x] Deploy to production new EC2 credentials and cleanup old ones.

**Related Issues/PRs**
- #1537

**Documentation**
- [Cleanup secrets](https://ui-team.global.cloud.sap/docs/house_keeping/secrets/)
- [Secrets documentation](https://ui-team.global.cloud.sap/docs/projects/elektra/secrets/)

PR #1537, opened 2025-03-04
research(core): Integrate Action Mailer with Limes Email API 

ArtieReus:
# Summary

This PR `showcases` how it could the limes Email API be integrated without having to redo the whole inquiry implementation. 

## TODOs
- Add to the `cloud_admin` the  `role:cloud_resource_admin or role:resource_service`
- Set the Limes Endpoint through a value

# Changes Made

- Implemented a global ApplicationMailer to enable inheritance and provide the `send_custom_email` method.
- Changed to test notification_new_project method from the inquiry

# Related Issues

- #1517

# Screenshots (if applicable)

Screenshot 2025-03-04 at 16 24 16

# Checklist

### DIFF
--- a/app/mailers/core_application_mailer.rb
+++ b/app/mailers/core_application_mailer.rb
@@ -4,7 +4,6 @@ require 'json'
 
 class CoreApplicationMailer < ActionMailer::Base
   layout "mailer"
-  API_URL = 'https://limes-campfire.qa-de-1.cloud.sap/v1/send-email'.freeze
 
   def send_custom_email(recipient:, subject:, body_html:)
     # Get the token form the cloud_admin instance
@@ -14,7 +13,7 @@ class CoreApplicationMailer < ActionMailer::Base
     token = cloud_admin.instance_variable_get(:@api_client).token
         
     # Set up the UI
-    uri = URI.parse(API_URL)
+    uri = URI.parse(Rails.configuration.limes_mailer_address)
     uri.query = URI.encode_www_form({ from: 'elektra' })
    
     # Set up the body for the request
--- a/app/mailers/core_application_mailer.rb
+++ b/app/mailers/core_application_mailer.rb
@@ -1,0 +1,56 @@
+require 'net/http'
+require 'uri'
+require 'json'
+
+class CoreApplicationMailer < ActionMailer::Base
+  layout "mailer"
+  API_URL = 'https://limes-campfire.qa-de-1.cloud.sap/v1/send-email'.freeze
+
+  def send_custom_email(recipient:, subject:, body_html:)
+    # Get the token form the cloud_admin instance
+    ###################################
+    # TODO: add to the cloud_admin user "role:cloud_resource_admin or role:resource_service"
+    ###################################
+    token = cloud_admin.instance_variable_get(:@api_client).token
+        
+    # Set up the UI
+    uri = URI.parse(API_URL)
+    uri.query = URI.encode_www_form({ from: 'elektra' })
+   
+    # Set up the body for the request
+    body = {
+      recipients: [recipient],
+      subject: subject,
+      mime_type: 'text/html',
+      mail_text: body_html
+    }.to_json
+
+    # Set up headers
+    headers = {
+      'Content-Type' => 'application/json',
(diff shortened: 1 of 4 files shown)


app/mailers/core_application_mailer.rb
```diff
@@ -0,0 +1,55 @@
...
+require 'uri'
+require 'json'
+
+class CoreApplicationMailer < ActionMailer::Base
+  layout "mailer"
+
+  def send_custom_email(recipient:, subject:, body_html:)
+    # Get the token form the cloud_admin instance
+    ###################################
+    # TODO: add to the cloud_admin user "role:cloud_resource_admin or role:resource_service"
```
andypf, line 11: What does it mean? Do we still have to do something?

plugins/inquiry/app/mailers/inquiry/inquiry_mailer.rb
```diff
@@ -81,7 +82,17 @@ def notification_new_project(inform_dl, inquiry, user_full_name)
...
+      
+      # Render the email body content
+      email_body = render_to_string('inquiry/inquiry_mailer/notification_new_project', layout: false)
+
+      send_custom_email(
+        recipient: inform_dl,
+        subject: subject,
+        body_html: email_body
+      )
+      # mail(to: inform_dl, subject: subject, content_type: "text/html")
```
andypf, line 94: Please remove this comment

plugins/inquiry/app/mailers/inquiry/inquiry_mailer.rb
```diff
@@ -81,7 +82,17 @@ def notification_new_project(inform_dl, inquiry, user_full_name)
           subject += "/#{@inquiry.tags["domain_name"]}"
         end
       end
-      mail(to: inform_dl, subject: subject, content_type: "text/html")
```
hgw77, line 84: is that the only place where we use the mailer?

andypf (commented):
Cool, I think there are still 3 methods using the `mail(to:`

app/mailers/core_application_mailer.rb
ArtieReus: We need to use the seeder to add the Dasboard user on of this roles:
```
role:cloud_resource_admin or role:resource_service
```
for ccadmin/cloud_admin

plugins/inquiry/app/mailers/inquiry/inquiry_mailer.rb
ArtieReus: Yes yes! I just leave it so you can see the difference

plugins/inquiry/app/mailers/inquiry/inquiry_mailer.rb
ArtieReus: No, we use it in several other places. This was just a single change I made to test the functionality. If we decide to proceed with this approach, we’ll need to update the other occurrences in the same file as well.

ArtieReus:
> mail(to::

if we go this way we need to replace the other methods also.

VoigtS:
Just for info: Campfire itself will be present in all prod regions (the rollout will happen during the upcoming days). The Cronus endpoint that we address is eu-de-1. So you can send requests to Campfire in the same region if you desire.

ArtieReus:
> Just for info: Campfire itself will be present in all prod regions (the rollout ~will happen during the upcoming days~ succeeded). The Cronus endpoint that we address is eu-de-1. So you can send requests to Campfire in the same region where elektra resides, if you desire.

These are great news, @VoigtS! When you say "in all prod regions," do you mean only the production regions where Cronus is deployed, or all 15 production regions?

ArtieReus:
@andypf If LimesMailAPI is available in all 15 production regions, we don't need to create an additional `cloud_admin` with a different identity URL. We can simply use the `cloud_admin` that is already being created for Elektra.

VoigtS:
@ArtieReus With every production region I mean every region where limes is deployed in. That's because we need to resolve project IDs to recipients in every region. Those regions should match with the regions where elektra is deployed.

hgw77 (approved):
well done :-)

PR #1541, opened 2025-03-12
chore(key-manager): make the 'new' UI default

andypf:
# Summary

Set the new Key Manager UI as the default option. The current interface still includes a note with a link to the old UI, allowing users to switch back if needed. Our goal is to gather more feedback from customers and address potential bugs. Eventually, we can consider removing the old UI completely.

# Checklist


hodanoori (approved):
Great, thanks Andreas

Issue #1536, opened 2025-03-04
[Task](identity): support keystone federation auth method


andypf:
## Task Description

As part of the **Bedrock** initiative, a new **federated authentication method** should be supported. For specific domains, users should see a **"Sign in"** button instead of the traditional login form. Clicking this button will redirect users to the **Keystone Federation** page, where they will be authenticated using the configured identity provider.

After successful authentication, Keystone will send a **POST request** to an endpoint specified by the `origin` parameter (from the **Sign in** button). This endpoint in **Elektra** should:

1. Accept the POST request.
2. Validate the provided **token** with Keystone.
3. Create a session on the Elektra side.
4. Redirect the user to the **domain's home page** on success or display an **error page** on failure.

## Sub-tasks

- [ ] Extend the **domain configuration** to distinguish between common login forms and the new **federation-based Sign in** page based on the domain name.
- [x] Create or extend an **authentication controller** with an action that:
    - Accepts POST requests from Keystone.
    - Uses the **monsoon_openstack_auth** gem to validate the provided **token**.
    - Creates a session if the token is valid.
    - Redirects the user (HTTP 302) to the domain's home page on success.
    - Shows an **error page** on failure, potentially redirecting back to the **Sign in** page.
- [x] Review and, if necessary, extend the **monsoon_openstack_auth** gem to support **token validation**.

## Acceptance Criteria

- [x] Federated authentication is configurable per domain.
- [x] Successful token validation creates a session and redirects the user.
- [ ] Invalid tokens display an error page.
- [ ] The **Sign in** button is only shown for configured domains.
- [ ] Existing login flow remains unchanged for non-federated domains.

- related to #1522

PR #1545, opened 2025-03-14
feat(core): add a controller to handle auth-token response

andypf:
# Summary

Add a new controller to handle auth-token response from keystone federation.

# Changes Made

- add new controller AuthTokenController
- create a new page views/auth_Token/verify to show verification errors
- sync with monsoon_openstack_auth gem

# Related to
- #1536 

# Checklist


app/views/auth_token/verify.html.haml
```diff
@@ -0,0 +1,7 @@
+
+.container
+
+  %section
+    .bs-callout.bs-callout-info.bs-callout-emphasize
+      Oops! We couldn't verify your token. Please ensure it's correct.
```
edda, line 6: We'll need to work on this error message a bit ;) It's not clear what the user should do here. I'm guessing trying to reauthenticate would be the way to go.

andypf: agree!

hgw77 (approved):

PR #1572, opened 2025-04-23
fix(dns-service): make search intuitive to use

### LINKED ISSUE #1203: Search in DNS view is unintuitive to use
When searching for a DNS zone the help text shows ``Search term can use wildcard "*".`` which suggests that you don't need to use wildcards to find something. In reality you need to pre- and suffix a search with wildcards to get the expected result.

For example if you search for ``global`` you would expect that to match ``global.example.com`` which is not the case unless you add the wildcards. Elektra should do that by default to make the search more end user friendly.

SuperSandro2000:
Closes #1203

### DIFF
--- a/plugins/dns_service/app/controllers/dns_service/zones_controller.rb
+++ b/plugins/dns_service/app/controllers/dns_service/zones_controller.rb
@@ -18,7 +18,7 @@ module DnsService
         if not params[:search].blank?
           @search = params[:search]
           @searchfor = "#{params[:searchfor]}"
-          # sourunding the user submitted search term with wildcards is stupid but the easiest way to make the search intuitive and match partial names
+          # surrounding the user submitted search term with wildcards is stupid but the easiest way to make the search intuitive and match partial names
           filter = { @searchfor.downcase() => "*" + @search + "*" }
         else
           params.delete(:search)
@@ -73,9 +73,9 @@ module DnsService
 
       if params.include?(:search)
         if not params[:search].blank?
-          @search = params[:search] 
+          @search = params[:search]
           @searchfor = "#{params[:searchfor]}"
-          # sourunding the user submitted search term with wildcards is stupid but the easiest way to make the search intuitive and match partial names
+          # surrunding the user submitted search term with wildcards is stupid but the easiest way to make the search intuitive and match partial names
           filter = { @searchfor.downcase() => "*" + @search + "*" }
         else
           params.delete(:search)
--- a/plugins/dns_service/app/controllers/dns_service/zones_controller.rb
+++ b/plugins/dns_service/app/controllers/dns_service/zones_controller.rb
@@ -19,7 +19,7 @@ module DnsService
           @search = params[:search]
           @searchfor = "#{params[:searchfor]}"
           # surrounding the user submitted search term with wildcards is stupid but the easiest way to make the search intuitive and match partial names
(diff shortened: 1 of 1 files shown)


plugins/dns_service/app/controllers/dns_service/zones_controller.rb
```diff
@@ -72,9 +73,10 @@ def show
 
       if params.include?(:search)
         if not params[:search].blank?
-          @search = params[:search]
+          @search = params[:search]
```
joker-at-work, line 76: this adds a space at the end of the line

plugins/dns_service/app/controllers/dns_service/zones_controller.rb
```diff
@@ -72,9 +73,10 @@ def show
 
       if params.include?(:search)
         if not params[:search].blank?
-          @search = params[:search]
+          @search = params[:search] 
           @searchfor = "#{params[:searchfor]}"
-          filter = { @searchfor.downcase() => @search }
+          # sourunding the user submitted search term with wildcards is stupid but the easiest way to make the search intuitive and match partial names
```
joker-at-work, line 78: same typo as above

joker-at-work (commented):
3 little nit-picks

plugins/dns_service/app/controllers/dns_service/zones_controller.rb
VoigtS: Thanks, this was a unnoticed leftover from testing.

plugins/dns_service/app/controllers/dns_service/zones_controller.rb
```diff
@@ -74,7 +75,8 @@ def show
         if not params[:search].blank?
           @search = params[:search]
           @searchfor = "#{params[:searchfor]}"
-          filter = { @searchfor.downcase() => @search }
+          # surrunding the user submitted search term with wildcards is stupid but the easiest way to make the search intuitive and match partial names
```
edda, line 78: Since we started nitpicking, this still isn't correct. A new typo appeared ;) 
```suggestion
          # surrounding the user submitted search term with wildcards is stupid but the easiest way to make the search intuitive and match partial names
```

edda (dismissed):

plugins/dns_service/app/controllers/dns_service/zones_controller.rb
```diff
@@ -18,7 +18,8 @@ def index
         if not params[:search].blank?
           @search = params[:search]
           @searchfor = "#{params[:searchfor]}"
-          filter = { @searchfor.downcase() => @search }
+          # surrounding the user submitted search term with wildcards is stupid but the easiest way to make the search intuitive and match partial names
+          filter = { @searchfor.downcase() => "*" + @search + "*" }
```
hgw77, line 22: it is better to use `"*#{@search}*"` it should work the same way as `"*" + @search + "*"`

plugins/dns_service/app/controllers/dns_service/zones_controller.rb
```diff
@@ -74,7 +75,8 @@ def show
...
```
hgw77, line 79: same as above, it is better to use `"*#{@search}*"` instead of `"*" + @search + "*"`

andypf (approved):
Super!

hgw77 (approved):

PR #1594, opened 2025-05-23
fix(identity): appcred list error handling

hgw77:
# Summary

*  this will adjust error handling for listing app credential to see the actual error message that is coming from the api

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -14,6 +14,9 @@ on:
       - synchronize
       - reopened
 
+permissions:
+  contents: read
+
 jobs:
   test:
     name: Run tests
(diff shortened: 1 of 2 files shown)


edda (approved):
lol, what kind of crazy error object is this? ;P

plugins/identity/app/javascript/widgets/app_credentials/List.jsx
```diff
@@ -48,7 +48,7 @@ const AppCredentialsList = ({ userId, refreshRequestedAt, projectId, overlay })
         setAppCredentialsFoundForProject(foundProject)
       })
       .catch((error) => {
-        setError(error.message)
+        setError(error.data.error.error.message)
```
edda, line 51: Actually with this deeply nested object you should do the conditional query, to not run into runtime errors if the object doesn't look like this and for good measure add a fallback error message:
```suggestion
        setError(error?.data?.error?.error?.message || "Something went wrong" )
```

edda (changes requested):

andypf (dismissed):

ArtieReus (dismissed):

andypf (approved):

ArtieReus (approved):

PR #1646, opened 2025-07-09
chore(core): adjust threads count for puma and add min threads

hgw77:
# Summary

* this will add the min threads option to the puma config and adjust default parameter
* with that we have more possibilities to fine tune the performance in different regions

# Checklist


andypf (approved):
👍

PR #1680, opened 2025-08-03
feat(plugins): add individual autoscaling configuration

VoigtS:
Castellum allows the configuration for either "All" share types or for "Individual" share types.
This PR adds this option to the UI.

# Changes Made

- Add selection between both configuration types.
- If a user swaps from "All" config to "Individual" config or vise versa with an active config, the swap will only be performed after the user agrees to reset the active configurations.
- If a user swaps from "All" config to "Individual" config or vise versa without an active config, the swap will be performed without asking the user to reset the existing configurations.
- Adjust autoscaling and asset events to the new structure.
- Fix autoscaling configuration modal. `max_size` is now a required field. It caused malformed API requests before.

# Screenshots (if applicable)

- Active config for all share types:
image

- Active individual configs:
image

# Checklist

### DIFF
--- a/plugins/shared_filesystem_storage/app/javascript/widgets/app/actions/castellum.js
+++ b/plugins/shared_filesystem_storage/app/javascript/widgets/app/actions/castellum.js
@@ -70,9 +70,7 @@ const fetchCastellumData =
 function filterShareTypeData(data = {}, isAllShares) {
   const matchingData = {}
   let allShares = isAllShares
-  if (isEmptyObject(data)) {
-    return { data: { "nfs-shares": null }, allShares }
-  }
+
   Object.keys(data).forEach((key) => {
     if (key === "nfs-shares") {
       allShares = true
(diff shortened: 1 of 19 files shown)


hgw77 (dismissed):
👍

majewsky (approved):

VoigtS:
Before merging, I had another look at it and noticed a weakness in the `asset scrape error` cache management. If autoscale configs are present and the initial scrape was performed, then the cache manager would always return the present data, regardless of whether autoconfigs for individual shareTypes were removed or added. Additionally, the collected shareIDs were not fetched properly before.

Both issues are fixed in: [976f9e4](https://github.com/SAP-cloud-infrastructure/elektra/pull/1680/commits/976f9e4320524064892d7adc2145abe331307157)

majewsky (approved):

hgw77 (approved):

Issue #1621, opened 2025-06-23
[Bug](Key Manager): downloaded keys are corrupted


hgw77:
**Description**

Using the new KeyManager UI, when downloading a secret, the result file contains additional double quotes `"` around the secret and all new lines within the secret are replaced by a text string `\r\n`.
This poses a potential problem, as the secret can no longer be used directly. One should remove the double quotes and replace `\r\n` with new lines. I tried saving the secret as different types - passphrase, private, certificate - the issue persists, no matter the type.

Accidentally I clicked to open the old UI and it turned out that the old UI downloads the secret without any modifications.

**Expected Behavior**

Keys are downloaded as it is

PR #1687, opened 2025-08-08
fix(key-manager): fixes download secret file formatting issue

hodanoori:
# Summary

This PR resolves a problem where downloading a secret resulted in a file with unwanted formatting changes. The downloaded file previously included additional double quotes (") around the secret. Moreover, all newlines within the secret were replaced by the literal string \r\n.

# Changes Made

- Write the raw secret payload directly to the file, without using` JSON.stringify `or any transformation that would change the content.

# Related Issues

- Issue 1: #1621 

# Screenshots (if applicable)

# Checklist

### DIFF
--- a/plugins/keymanagerng/app/javascript/widgets/app/components/secrets/secretDetails.jsx
+++ b/plugins/keymanagerng/app/javascript/widgets/app/components/secrets/secretDetails.jsx
@@ -102,10 +102,8 @@ const SecretDetails = () => {
         setPayloadString(data)
       }
       if (payloadRequested) {
-        //Downloading the payload content as a file
-        const fileData =
-        typeof data === "string" ? data : String(data) // Use raw string
-        const blob = new Blob([fileData], {
+        //Downloading the payload content as a file, put the raw data into a blob
+        const blob = new Blob([data], {
           type: secret?.data?.content_types?.default,
         })
         const url = URL.createObjectURL(blob)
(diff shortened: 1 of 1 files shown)


hgw77 (approved):

plugins/keymanagerng/app/javascript/widgets/app/components/secrets/secretDetails.jsx
```diff
@@ -103,7 +103,8 @@ const SecretDetails = () => {
       }
       if (payloadRequested) {
         //Downloading the payload content as a file
-        const fileData = JSON.stringify(data)
+        const fileData =
```
hgw77, line 106: is this right? there is no value

hodanoori: yes, it is. Where there is no value? I tested it locally and the same payload as you see in the secret details, will be inside the downloaded file

hodanoori: I check if the type of data is string, then I do not need to use String() function to convert it to string, is your question answered?

plugins/keymanagerng/app/javascript/widgets/app/components/secrets/secretDetails.jsx
```diff
@@ -103,7 +103,8 @@ const SecretDetails = () => {
...
+        typeof data === "string" ? data : String(data) // Use raw string
```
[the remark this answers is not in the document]
hodanoori: Good point, thanks for the hint. There are couple of binary types for payload like `"application/pkcs8"`, `"application/pkix-cert"`, `"application/octet-stream"` which converting them into string corrupt the file format. Therefore I just use the raw data

PR #2007, opened 2026-04-15
feat(identity): improve rescope error handling with manual rescoping

andypf:
## Overview

This PR improves upon #2006 by refactoring how we handle rescoping errors. Instead of relying on `rescue_from` blocks, we now handle rescope failures directly in a `before_action` callback, giving us better control over the error handling flow.

## Problem with Previous Approach

In #2006, we used `rescue_from MonsoonOpenstackAuth::Authentication::NotAuthorized` to catch rescope failures. However, this approach had limitations:

- The exception was caught **after** it was raised, interrupting the normal request flow
- For domain-scoped requests, we couldn't easily continue to the controller action
- Returning `nil` from the rescue block caused blank pages
- Error handling logic was split between the rescue block and the actual rescoping

### Manual Rescoping with Error Handling

```ruby
authentication_required(
  # ... other options ...
  rescope: false  # Disable automatic rescoping
)

before_action :rescope_token_with_error_handling
```

The new `rescope_token_with_error_handling` method:
1. Calls `authentication_rescope_token` manually
2. Catches `NotAuthorized` exceptions inline
3. Handles errors based on scope type:
   - **Project-scoped**: Renders unauthorized/not_found with HTTP 200
   - **Domain-scoped**: Logs and continues (allows action to execute)

### Simplified Rescue Concern

Also cleaned up `app/controllers/concerns/rescue.rb`:
- Removed the `else` branch that redirected to login
- Now always renders unauthorized template with HTTP 200
- Simpler, more consistent error handling

## Benefits

✅ **No more blank pages** - domain-scoped requests continue normally  
✅ **Cleaner error handling** - errors caught where they occur  
✅ **Better control** - can decide per-scope how to handle failures  
✅ **Prevents OAuth loops** - HTTP 200 for authorization failures  
✅ **Maintains #2006 fixes** - all previous improvements still work  

## Testing

All 342 tests passing:
```bash
docker exec elektra-dev bundle exec rspec spec/controllers/dashboard_controller_spec.rb plugins/monsoon-openstack-auth/spec/
```

## Changes

- `app/controllers/dashboard_controller.rb`: Added manual rescoping with `rescope: false` and `rescope_token_with_error_handling`
- `app/controllers/concerns/rescue.rb`: Simplified NotAuthorized handling

## Related

- Builds on #2006 (OAuth redirect loop fix)

### DIFF
--- a/app/controllers/concerns/rescue.rb
+++ b/app/controllers/concerns/rescue.rb
@@ -53,20 +53,12 @@ module Rescue
       # "MonsoonOpenstackAuth::Authentication::NotAuthorized" are rescued directly in rescope_token and handled by
       # "rescue_and_render_exception_page" but I leave it here just in case ;-)
       rescue_from "MonsoonOpenstackAuth::Authentication::NotAuthorized" do |exception|
-        # for project "has no access to project"
-        # check MonsoonOpenstackAuth/Authentication/auth_session.rb
-        if exception.message =~ /has no access to project/
-          # Return 200 OK instead of 401/403 to prevent OAuth redirect loop
-          # User IS authenticated (passed OAuth), just not authorized for this scope
-          # Using 401 would cause OAuth proxy to redirect back to IdP, creating a loop
-          render(template: "application/exceptions/unauthorized")
-        else
-          redirect_to monsoon_openstack_auth.login_path(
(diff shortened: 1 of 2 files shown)


edda (approved):

hgw77 (approved):
super! :-)

PR #2018, opened 2026-04-24
fix(monsoon-openstack-auth): clear stale tokens and improve scope validation

andypf:
## Problem

After cookie-session migration, expired tokens in the session caused an infinite loop where users couldn't see auth_projects:

1. Token expires but remains in session storage (`session[:auth_token_value]`)
2. `validate_token` returns `nil` for expired token
3. Session is **not cleared**, so next request tries the same expired token again
4. User sees no `auth_projects` and cannot access resources
5. Only manually deleting **ALL cookies** fixed the issue

Additionally, the scope validation was using URL parameters (friendly IDs) which required extra `FriendlyIdEntry` lookups and was error-prone.

## Root Cause

When migrating from database sessions to cookie sessions (PR #1925), the token storage changed:
- **Before**: Token stored in database via `TokenStore` with complex expiry handling
- **After**: Token value stored directly in session cookie (`session[:auth_token_value]`)

The problem: When a token expired, `validate_token` returned `nil`, but the expired token value remained in the session, causing an infinite authentication loop.

### 1. Clear stale tokens on validation failure

When `validate_token` returns `nil` (expired/invalid token):
- Clear `session[:auth_token_value]` if session was the token source
- Delete cross-dashboard cookie if it was the source
- This breaks the infinite loop and forces re-authentication on next request

When `validate_token` throws an exception:
- Clear both session token and cross-dashboard cookie
- Log error with source information for debugging

### 2. Improved scope validation with `token_domain_matches_scope_domain?`

Renamed and refactored from `token_domain_matches_url?`:
- Uses `@scope` (already resolved domain/project IDs) instead of URL params
- No dependency on `FriendlyIdEntry` lookups (redundant, already resolved in `@scope`)
- Proper validation of both project and domain scopes:
  - **Project scope**: checks both project ID and domain ID match
  - **Domain scope**: checks domain ID matches (token must not be project-scoped)
  - Handles both domain-scoped and project-scoped tokens correctly

## Testing

Added 5 new test scenarios for expired token handling:
- Expired session token → session cleared
- Expired cross-dashboard cookie token → cookie deleted
- Both cookies with same expired token → both cleared
- Token validation exception (e.g., `Excon::Errors::Unauthorized`) → cleanup
- Scope mismatch tests updated to use `@scope` instead of URL params

All existing tests pass.

## Impact

✅ Users will no longer experience the "no auth_projects" issue with expired tokens
✅ Automatic session cleanup ensures proper redirect to login when tokens expire
✅ More reliable scope validation without FriendlyIdEntry dependency
✅ Better logging for debugging token validation issues

## Checklist

- [x] Root cause identified and documented
- [x] Fix implemented with proper error handling
- [x] Tests added for all new scenarios
- [x] Logging improved for debugging
- [x] No breaking changes to existing functionality

### DIFF
--- a/plugins/monsoon-openstack-auth/lib/monsoon_openstack_auth/authentication/auth_session.rb
+++ b/plugins/monsoon-openstack-auth/lib/monsoon_openstack_auth/authentication/auth_session.rb
@@ -321,14 +321,14 @@ module MonsoonOpenstackAuth
           token = @api_client.validate_token(auth_token_value)
 
           if token
-            # Check if token's domain matches URL's domain
-            unless token_domain_matches_url?(token)
-              MonsoonOpenstackAuth.logger.info "Token domain mismatch with URL (source: #{token_source}), rejecting token." if @debug
+            # Check if token's domain matches the requested scope
+            unless token_domain_matches_scope_domain?(token)
+              MonsoonOpenstackAuth.logger.info "Token scope mismatch (source: #{token_source}), rejecting token." if @debug
 
               # If the mismatched token came from cross-dashboard cookie, delete it
               # This prevents repeated authentication failures with stale cookies
               if token_source == :cross_dashboard_cookie
-                MonsoonOpenstackAuth.logger.info 'Deleting stale cross-dashboard cookie due to domain mismatch.' if @debug
+                MonsoonOpenstackAuth.logger.info 'Deleting stale cross-dashboard cookie due to scope mismatch.' if @debug
                 self.class.delete_cross_dashboard_cookie(@controller)
               end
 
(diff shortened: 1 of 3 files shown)


andypf:
## Update

Code has been reviewed and is ready for merge. 

**Key improvements in this version:**
- ✅ Stale token cleanup properly handles all edge cases (session, cookie, both)
- ✅ Improved scope validation using `@scope` instead of URL parameters
- ✅ Comprehensive test coverage for expired token scenarios
- ✅ Better error logging with token source information

**Root cause confirmed:**
The "Unauthorized" issue when switching between domains was caused by:
1. Expired tokens remaining in session storage (infinite loop)
2. FriendlyID-based domain validation being error-prone during domain switches

Both issues are now resolved. Users will be automatically redirected to login when tokens expire, and domain switching will work correctly with the improved scope validation.

hgw77 (approved):

edda (approved):

PR #2024, opened 2026-04-28
feat(core): prevent oauth2-proxy redirect loops with middleware

andypf:
## Problem

When oauth2-proxy sits in front of Elektra (configured in nginx-ingress), it intercepts 401 and 403 HTTP status codes and treats them as "user not authenticated", triggering a redirect to the SSO provider. This creates endless redirect loops.

This issue became more prominent after the cookie session migration in March 2026, which introduced cross-dashboard authentication and increased token validation scenarios that can result in 401/403 responses.

## Root Cause

Many controllers across the codebase render 401/403 status codes:
- 79 locations across 16 plugin files (`kubernetes_ng`, `block_storage`, `lbaas2`, etc.) use `status: e.code` which returns the raw OpenStack error code
- Authentication and authorization failures naturally return 401/403
- These status codes trigger oauth2-proxy to assume the user is not authenticated

## Solution

This PR implements a **Rack middleware** (`OauthProxyProtectionMiddleware`) that:

1. **Intercepts all 401/403 responses** from the Rails application
2. **Converts them to 200 OK** before they reach oauth2-proxy
3. **Preserves the original response body** completely unchanged
4. **Logs each interception** for debugging purposes
5. **Only runs in production/development** (disabled in test environment to allow tests to verify actual status codes)

The middleware is positioned in the Rack stack to run AFTER the Rails app generates responses but BEFORE they reach nginx-ingress/oauth2-proxy.

## Why This Approach?

This middleware solution provides:

- ✅ **Global coverage** - automatically protects all controllers and plugins without individual changes
- ✅ **Future-proof** - new plugins automatically get protection
- ✅ **No breaking changes** - response bodies remain unchanged; clients can still detect errors
- ✅ **Clean architecture** - single point of responsibility, no scattered controller overrides
- ✅ **Test-friendly** - disabled in test environment so tests can verify actual HTTP status codes

## Changes

1. **Added `OauthProxyProtectionMiddleware`** - new Rack middleware in `app/middleware/`
2. **Registered middleware** in `config/application.rb` with `unless Rails.env.test?` guard
3. **Removed duplicate `rescue_from`** in `DashboardController` (already handled by Rescue concern)
4. **Added comprehensive tests** - 14 RSpec examples covering all middleware behavior

## Testing

All tests pass:
- ✅ Middleware specs (14 examples)
- ✅ DashboardController specs (18 examples)
- Middleware correctly disabled in test environment (verified)

## Deployment Notes

After deployment, monitor logs for:
```
OauthProxyProtectionMiddleware: Intercepted 401/403 response for [METHOD] [PATH]
```

These log entries indicate the middleware is actively preventing redirect loops.

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -22,12 +22,6 @@ class DashboardController < ::ScopeController
   before_action :set_mailer_host, except: %i[terms_of_use]
   before_action :load_help_text, except: [:terms_of_use]
 
-  # Handle case when user is not authenticated at all
-  rescue_from MonsoonOpenstackAuth::Authentication::NotAuthenticated do |exception|
-    # User is not logged in - return 401 to trigger OAuth login
-    render template: 'application/exceptions/not_authenticated', status: :unauthorized
-  end
-
   def check_terms_of_use
     @orginal_url = request.original_url
     return if tou_accepted? || @domain_config&.feature_hidden?('terms_of_use')
--- a/app/middleware/middlewares.rb
+++ b/app/middleware/middlewares.rb
@@ -7,6 +7,7 @@ require_relative "inquiry_metrics_middleware"
 require_relative "http_metrics_collector_middleware"
 require_relative "sli_metrics_middleware"
 require_relative "http_metrics_exporter_middleware"
+require_relative "oauth_proxy_protection_middleware"
 
 module Middlewares
 end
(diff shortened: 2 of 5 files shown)


ArtieReus (approved):
Tested locally and seems to work 🎉. Thanks!!
UI adjustments will be created in an extra PR

hgw77 (approved):

edda (approved):

PR #2055, opened 2026-05-19
feat(monsoon-openstack-auth): add browser-side mTLS SSO login via JS

bbobrov:
Implement a new authentication flow where the browser directly authenticates with Keystone using the cc_x509 external plugin via mTLS (client certificate presented through the upstream ingress), then posts the resulting token to Elektra's consume-auth-token endpoint to establish a Rails session.

Changes:
- Add sso_login.ts module that auto-fires on the login page, performing a cross-origin fetch to Keystone with X-User-Domain-Name header and methods=['external'], then posting the X-Subject-Token back to Elektra same-origin with CSRF protection
- Add esbuild entry point (app/javascript/sso_login.ts)
- Add meta tags to the login view for keystone-url, domain, and consume-path (only rendered when MONSOON_OPENSTACK_AUTH_API_ENDPOINT is configured)
- Extend SessionsController#consume_auth_token to return JSON for raw=1 POST requests (the browser SSO path) while preserving existing redirect behaviour for legacy flows

The legacy server-side validate_sso_certificate path is preserved unchanged as a fallback.

The template below is left empty on purpose, nothing from it was done.

# Summary

# Changes Made

- Change 1
- Change 2
- Change 3

# Related Issues

- Issue 1: #ISSUE_ID

# Screenshots (if applicable)

# Checklist


hgw77:
this can be close, @andypf is working on that topic

PR #2064, opened 2026-05-22
fix(compute): hide Lock/Unlock menu options 

### LINKED ISSUE #1666: [Bug](Compute): "Lock" option still visible in menu for already locked server instances
**Description**

When a server instance is in the locked state, the UI correctly displays the lock icon. However, the Lock option is still available in the action menu alongside Unlock, which is misleading. The Lock option should be hidden or disabled when the instance is already locked.

**Steps to Reproduce**

1. Choose Compute --> Servers.
2. Lock a server instance.
3. Observe the lock icon next to the instance.
4. Open the action menu for that instance.
--> Both **Lock** and **Unlock** options are visible, even though the instance is already locked.

**Expected Behavior**

Only the **Unlock** option should be shown in the menu when a server instance is locked.

**Screenshots**

Image

**Environment**

- OS: Windows
- Browser: Chrome
- Version: 

**Additional Context**

EdrilanBerisha:
Fixes https://github.com/SAP-cloud-infrastructure/elektra/issues/1666

When a server instance was already locked, both Lock and Unlock options were visible in the action menu, which was misleading. The root cause was that the visibility of Lock/Unlock was controlled only by user permissions (compute:instance_lock / compute:instance_unlock), without checking the actual locked state of the instance.

Fix: add instance.locked? checks to the action menu conditions in plugins/compute/app/views/compute/instances/_item_actions.html.haml

  Before:
    - if current_user.is_allowed?("compute:instance_lock") show Lock link
    - if current_user.is_allowed?("compute:instance_unlock") show Unlock link

  After:
    - if current_user.is_allowed?("compute:instance_lock") && !instance.locked? show Lock link (only when NOT locked)
    - if current_user.is_allowed?("compute:instance_unlock") && instance.locked? show Unlock link (only when IS locked)

This ensures:
- Only Lock is shown when the instance is NOT locked
- Only Unlock is shown when the instance IS locked
- They are mutually exclusive and never both visible at the same time

Also added spec/models/server_lock_spec.rb with 30+ tests covering:
- locked?() with all truthy forms (true, 'true' via metadata or top-level)
- locked?() with all falsy forms (false, 'false', nil, not set)
- Action menu visibility conditions for both locked and unlocked states
- Mutual exclusivity invariant across all locked attribute values
- Metadata vs top-level locked attribute precedence

hey @hgw77 can you review this here?


hgw77 (approved):
👍 good job

EdrilanBerisha:
as I need 2 reviewers to merge this I would ask you @edda could you help and review this PR?

Danke!

ArtieReus (dismissed):

plugins/compute/spec/models/server_lock_spec.rb
hgw77, line 1: @EdrilanBerisha did you run the rspec tests locally because these tests failing in our Rails test action

I investigated the problem a little and The RSpec tests are failing because the `locked?` method in `Compute::Server` doesn't handle nil cases properly:

When no metadata is passed, metadata returns nil, causing: `NoMethodError: undefined method 'locked' for nil:NilClass`

When metadata hash is passed but `@service` is `nil` (in tests), attribute_to_object fails: `NoMethodError: undefined method 'map_to' for nil:NilClass`

Current [implementation](https://github.com/SAP-cloud-infrastructure/elektra/blob/master/plugins/compute/app/models/compute/server.rb#L503-L505)

````
  def locked?
    metadata.locked == true || metadata.locked == "true" || locked == true || locked == "true"
  end
````

possible Fix:

````
  def locked?
    metadata_locked = (metadata&.locked rescue nil)
    top_level_locked = read("locked")

    metadata_locked == true || metadata_locked == "true" ||
    top_level_locked == true || top_level_locked == "true"
  end
````

This uses rescue to catch any errors from the metadata call so and ensures the method works both in production (with a service) and in tests (without a service), and handles the case when no metadata is present at all.

**BUT:**

These unit tests for locked? don't add value and are forcing changes to production code that already works.
In other words, the locked? method already exists and works in production - it doesn't need these unit tests.

So I suggest, just delete the new rspec test

hgw77 (changes requested):

plugins/compute/spec/models/server_lock_spec.rb
EdrilanBerisha: thank you for pointing out, removed the rspec test, back to you :)

hgw77 (approved):

andypf (approved):

PR #2123, opened 2026-07-22
feat(monsoon-openstack-auth): add SSO precheck with deployment-level control

andypf:
## Summary
Implements automatic SSO login attempt before showing the login form, with per-deployment configuration via `SSO_PRECHECK_ENABLED` environment variable.

## Features
- Frontend SSO precheck with spinner and 5-second timeout
- Graceful fallback to manual login if SSO unavailable
- Clean separation of concerns: views in root app, styles in `_login.scss`
- Feature toggle: `SSO_PRECHECK_ENABLED=true` enables frontend precheck

## Changes
- Add SSO precheck JavaScript with timeout and error handling
- Extract login form into reusable partial (`_login_form.html.haml`)
- Add `.sso-precheck-loader` CSS for spinner alignment
- Add `keystone_tokens_url` helper method in sessions controller
- Add `SSO_PRECHECK_ENABLED` env var to toggle between:
  - Frontend SSO precheck (modern, certificate-based auto-login)
  - Backend SSL cert auth (classic method)

## Architecture
- **Views**: `/app/views/monsoon_openstack_auth/sessions/` (override pattern)
- **Styles**: `/app/assets/stylesheets/_login.scss` (root app)
- **JavaScript**: `/plugins/monsoon-openstack-auth/app/javascript/sso_precheck.js`
- **Compiled**: `/app/assets/builds/monsoon-openstack-auth_plugin.js` (esbuild)

## Testing
- Set `SSO_PRECHECK_ENABLED=true` to enable SSO precheck
- Set `SSO_PRECHECK_ENABLED=false` for traditional backend SSL cert auth
- Spinner shows "Checking for SSO session..." during precheck
- Timeout after 5 seconds with fallback to login form

### Check SSO
image

### Create Session
image

## Checklist
- [x] Code follows Elektra design principles
- [x] Clean separation: root app views/styles, plugin JavaScript
- [x] Feature toggle allows per-deployment control
- [x] Graceful error handling and fallback
- [x] No breaking changes to existing auth flows


app/views/monsoon_openstack_auth/sessions/_login_form.html.haml
```diff
@@ -0,0 +1,61 @@
+- sso_precheck_enabled = ENV['SSO_PRECHECK_ENABLED'].to_s == 'true'
+
+- if sso_precheck_enabled
+  #sso-precheck-container
+    .sso-precheck-loader
+      .spinner
+      Checking for SSO session...
+
+#login-content{ style: sso_precheck_enabled ? "display: none" : "" }
```
edda: Using only css to hide the form still makes it accessible very easily if you know your way around browser dev tools. Why not use haml like above? `- if sso_precheck_enabled`

andypf: Good point! The reason we render the login form even when SSO precheck is enabled is that JavaScript needs to show it as a fallback if the SSO check fails (timeout, no valid session, network error, etc.).

andypf: but I changed from inline styles to css class

hgw77 (approved):

app/views/monsoon_openstack_auth/sessions/_login_form.html.haml
```diff
@@ -0,0 +1,66 @@
+- sso_precheck_enabled = ENV['SSO_PRECHECK_ENABLED'].to_s == 'true'
+
+- if sso_precheck_enabled
+  #sso-precheck-container
+    .sso-precheck-loader
+      .spinner
+      Checking for SSO session...
+
+#login-content{ class: ("login-hidden" if sso_precheck_enabled) }
```
edda, line 9: You still only hide the form via CSS. Which means people could use it even if sso_precheck_enabled. Why not conditionally render it using haml?


```suggestion
- unless sso_precheck_enabled
    #login-content
```

edda: nvm, missed the earlier comment

edda (approved):

Issue #2133, opened 2026-08-10
[Bug](prodel): Deletion of project with child projects doesn't surface backend error


stanislav-zaprudskiy:
**Description**

Deleting a project that has child projects fails in prodel with HTTP 403, but is swallowed in Elektra returning home page with HTTP 200.

**Steps to Reproduce**

1. Create a project and a child project.
2. Try to delete the parent project.

**Expected Behavior**

The deletion fails, surfacing prodel (or decorated) error.

```
{
  "code": 403,
  "errors": {
    "openstack.exceptions.ForbiddenException": "ForbiddenException: 403: Client Error for url: https://..., Cannot delete the project ${ID} since it is not a leaf in the hierarchy. Use the cascade option if you want to delete a whole subtree."
  },
  "message": "ForbiddenException: 403: Client Error for url: https://identity-3.${OS_REGION_NAME}.cloud.sap/v3/projects/${ID}, Cannot delete the project ${ID} since it is not a leaf in the hierarchy. Use the cascade option if you want to delete a whole subtree.",
  "status": "Forbidden"
}
```

Note: cascade deletion is not supported.

**Additional Context**

The current behavior is very confusing, giving a user a false impression that project has been deleted, making users to wonder then, why is it re-appearing (though it's never been deleted in the first place).

PR #2136, opened 2026-08-19
fix(identity): improve error handling and loading states for prodel

hgw77:
# Summary

Fix silent failure when deleting a project with child projects. The backend returns HTTP 403 with a clear error message, but the UI was not surfacing it to the user.

# Changes Made

- Extract error message string in `.catch()` instead of storing the raw error object ( `errorData?.error?.message || errorData?.message || ...`)
- Move the <Message> error component outside the {data && !loading} block so it renders regardless of fetch state
- Replace the ternary data ? ... : `<p>Loading...</p>` with independent conditions so a fetch error does not show "Loading..." instead of the error
- Clear apiError at the start of each delete attempt (setApiError(null))
- Remove leftover console.log/console.error debug calls

## Flow Diagram

```
  DELETE /os-api/prodel/api/v1/projects/:id
    │
    ▼
  OsApiController#reverse_proxy
    │
    └─ elektron → Prodel Backend
                    │
                    └─ HTTP 403
                       {"message": "Cannot delete the project since
                        it is not a leaf in the hierarchy..."}
    │
    ▼
  rescue → render json: { error: { message: "..." } }, status: 403
    │
    ▼
  ajax_helper.js  (response.ok === false → throw Error)
    error.data = { error: { message: "..." } }
    │
    ▼
  onConfirm() .catch()

    BEFORE                          AFTER
    ──────────────────────────      ──────────────────────────
    setApiError(error.data)         message = error.data?.error?.message
    ↓                               setApiError(message)
    {data && !loading ?             ↓
      <Message text={               {apiError &&
        apiError.error.message}/>     <Message text={apiError}/>}
    : <p>Loading...</p>}            ← always visible, independent of data
    ↑
    never rendered when data===null
```
  
# Related Issues

  - Issue: #2133

 # Checklist


ArtieReus (commented):
Don't we want to add a test for the fixes?

hgw77:
good catch, will take a look into that

andypf (approved):

TilmanHaupt (approved):

PR #2151, opened 2026-09-03
feat(dns-service): display only available DNS pools in zone request wizard

### LINKED ISSUE #2144: [Task](dns): Display only available DNS Pools in DNS request wizard
**Task Description**

Elektra to display only DNS Pools which are available in the current Keystone Domain during DNS Zone request.

**Additional Context**

Currently, when a user doesn't have sufficient role to create a DNS zone it should be requested via the "DNS wizard'.
As we implemented Pool RBAC in Designate we can now control which Keystone Domain has access to which DNS Pool. Users who don't have access to specific DNS Pool should not be able to request or create a zone in it. However, in the Elektra UI the user is able to request DNS zone creation in any DNS Pool available in the region. Presumably, because Elektra fetches the list of Pools with cloud admin privileges.

Instead, we should display only the Pools which are allowed (were shared) to the currently active Keystone domain.

To see which Pool is shared to which Keystone domains one could use our (patched) Openstack CLI:
`openstack dns pool list `as a cloud admin
`openstack dns pool share list DOMAIN_UUID` as a cloud admin

On the image below, all regional Pools are visible, where in fact only the first one should be visible due to current user context (domain):

Image

hgw77:
# Summary

Filters the DNS pool list in the zone request wizard so that users only see
pools that have been shared to their current Keystone domain. Previously, all
pools in the region were shown regardless of whether the user's domain had
access to them.

# Changes Made

- Added `pool_shares(pool_id)` service method that calls
  `GET /v2/pools/{pool_id}/shares` (with cloud admin privileges) and returns
  the list of Keystone domains the pool is shared with
- Added `pools_for_domain(domain_id)` service method that fetches all pools,
  queries their shares, and returns only the ones where `target_domain_id`
  matches the current scoped domain
- Updated `RequestZoneWizardController#new` and `#create` (error re-render) to
  use `pools_for_domain(@scoped_domain_id)` instead of `pools[:items]`
- Falls back to showing all pools if the shares API is unavailable, to avoid
  breaking environments where pool RBAC is not configured

# Related Issues

- Issue 1 closes: #2144

# Checklist


andypf (approved):
Great!

ArtieReus (approved):

PR #2154, opened 2026-09-07
fix(metrics): send maia auth token via POST body instead of URL

wiltongarcia:
## Summary

- Replaces the \"Open Maia Dashboard\" `<a>` link with an HTML POST form so the OpenStack identity token (`x-auth-token`) is sent in the request body instead of the URL query string
- Removes token exposure from browser URL bar, history, server access logs, and referrer headers
- Updates the E2E test selector from `a:has-text` to `button:has-text` to match the new element

## Security motivation

The previous pattern embedded the live Keystone token directly in the URL:
```
https://maia.<region>.cloud.sap/<domain>?x-auth-token=<TOKEN>
```
This made the token visible in browser history, server access logs on the Maia side, and any HTTP referrer headers on subsequent navigations.

## Dependency

Maia must accept `POST /:domain` with `x-auth-token` in the form-encoded body and apply a POST/Redirect/GET pattern to keep the final URL clean. A companion change is required on the [Maia](https://github.com/SAP-cloud-infrastructure/maia) side before this can go to production.

## Test plan

- [ ] Confirm Maia accepts `POST /:domain` with `x-auth-token` in body (coordinate with Maia team)
- [ ] Manually verify "Open Maia Dashboard" button opens Maia in a new tab and authenticates correctly
- [ ] Run E2E test: `pnpm e2e:ui -- --host http://localhost:PORT metrics-member-functional`
- [ ] Confirm token no longer appears in the browser URL or network request query string


plugins/metrics/app/views/metrics/application/index.html.haml
```diff
@@ -29,8 +29,10 @@
           series and perform ad-hoc queries leveraging all of PromQL's rich query syntax.
         %p
 %p
-- maia_dashboard_url = "https://maia.#{current_region}.cloud.sap/#{@scoped_domain_name}?x-auth-token=#{current_user.token}"
```
hgw77, line 32: the token is no longer needed or is that remove by accident?

hgw77 (dismissed):

plugins/metrics/app/views/metrics/application/index.html.haml
hgw77: oh, I see, I did not read the manual ;-)

hgw77 (approved):

TilmanHaupt (approved):

